Issue #2, opened 2015-06-24
Merge content of preboot as a single commit


tbosch:

gdi2290:
closed via https://github.com/angular/universal/pull/7

PR #7, opened 2015-07-06
Add preboot

jeffwhelpley:
This works for me in my preboot_basic example but I have not integrated it yet with server/client yet. I talked to Patrick and we decided it would be easiest at this point to check in what I have and he will handle integrating it the rest of the way.


jeffwhelpley:
fixes #2

PR #9, opened 2015-07-10
fix(minor): Fixes for getClientCode and normalize

jeffwhelpley:
Includes several fixes on the server side which checks input options from the user


gdi2290:
merged via https://github.com/angular/universal/commit/20c4d9a64f78d09aa6f4cc726898756bfe5f9b07

Issue #64, opened 2015-08-07
refactor batter out and use gulp in gulpfile 


gdi2290:
configuration is not clear. We have configuration living in task file and gulpfile. Having configurations in place place would allow for simpler reasoning on each moving part. Having a task per file limits the ability to compose multiple tasks. The repo doesn't need this level of abstraction just yet and would benefit more by having one gulpfile either requiring the configuration or keeping it inline

manekinekko:
@gdi2290 do you need me to move all gulp tasks into one single gulpfile ?

manekinekko:
I am on it.

gdi2290:
@manekinekko yeah all in one gulpfile

manekinekko:
PR sent (https://github.com/angular/universal/pull/69)

gdi2290:
fixed via https://github.com/angular/universal/pull/69

PR #69, opened 2015-08-09
chore(build): refactor batter out and use gulp in one single gulpfile

manekinekko:
Task https://github.com/angular/universal/issues/64

Notes: 
 1. Batter is out.
 2. Still need to figure out what should **play** and **play.preboot** do.


manekinekko:
Travis is unhappy because there is a missing semicolon in here [server_dom_renderer.ts#L26](https://github.com/angular/universal/blob/master/modules/server/src/server_dom_renderer.ts#L26)  ^^

gdi2290:
thanks! :+1:

Issue #50, opened 2015-08-05
Create NodeBackend for Http


gdi2290:
This is really easy to do since you could just include a Node version of XMLHttpRequest (called NodeXhr) that replaces BrowserXhr for an MVP

in this file [modules/server/src/http_server.ts](https://github.com/angular/universal/blob/master/modules/server/src/http_server.ts)

replace [XMLHttpRequest ](https://github.com/angular/angular/blob/master/modules/angular2/src/http/backends/browser_xhr.ts#L7) with a node version [xmlhttprequest](https://www.npmjs.com/package/xmlhttprequest) as `NodeXhr`

and reuse the same `XHRBackend` with `bind(ConnectionBackend)
      .toClass(XHRBackend)` rather than `NodeBackend` and that should work

manekinekko:
I'll take care of this one.

gdi2290:
Here's the code currently for it.
https://github.com/angular/universal/blob/master/modules/server/src/http_server.ts
at the moment it's stubbed out to use the Mock version.

manekinekko:
@gdi2290  Can you give more details about what you expect from this? The issue description lacks some details for new contributors (like me).

gdi2290:
@manekinekko I updated the original post. Ping me again if you need more details/help

manekinekko:
thanks @gdi2290. I will let you know. I need to take some time and wrap my head around how preboot works under the hood.

manekinekko:
@gdi2290 do you really want to go with [node-XMLHttpRequest](https://github.com/driverdan/node-XMLHttpRequest), it seems to have some [limitations](https://github.com/driverdan/node-XMLHttpRequest#known-issues--missing-features). I dont know what level of support we want here.

I've found other [alternatives](https://www.npmjs.com/search?q=xmlhttprequest) such as [node-xhr2](https://github.com/pwnall/node-xhr2) which implements obviously XHR2 specs. There is also the well known [request](https://github.com/request/request) module.

gdi2290:
@manekinekko `node-xhr2` looks great! I only picked `node-XMLHttpRequest` as a suggestion (first one, no research)

gdi2290:
[...]

Issue #44, opened 2015-08-04
Environment or Server service for client/server


gdi2290:
A service for server concerns such as but not limited to SEO. 
related interface proposals https://github.com/angular/angular/issues/3144#issuecomment-122698339 
Either `Environment` that's bound by di to `Server` or `Client`. There is also an optimization step per environment that will be solved via Webpack build
prior art https://github.com/DerMambo/ms-seo

```es6
var robotTxt = `
  User-agent: Twitterbot
  Disallow:

  User-agent: *
  Disallow: /
`;

var metatags = `
  <meta name="twitter:card" content="summary" />
  <meta name="twitter:site" content="@nytimesbits" />
  <meta name="twitter:creator" content="@nickbilton" />
  <meta property="og:url" content="http://bits.blogs.nytimes.com/2011/12/08/a-twitter-for-my-sister/" />
  <meta property="og:title" content="A Twitter for My Sister" />
  <meta property="og:description" content="In the early days, Twitter grew so quickly that it was almost impossible to add new features because engineers spent their time trying to keep the rocket ship from stalling." />
  <meta property="og:image" content="http://graphics8.nytimes.com/images/2011/12/08/technology/bits-newtwitter/bits-newtwitter-tmagArticle.jpg" />
`;

@Component()
@View()
@RouteConfig()
class App {
  constructor(server: Server) {
    server.set('title', 'My home page');
    server.set('robot.txt', robotTxt);
    server.set('metatags', metatags);
  }
}
```
```es6
@Component({
  selector: 'app',
  bindings: [ bind(Environment).toClass(Server) ]
})
@View({
})
class App {
  constructor(env: Environment) {
  }
}
```

the idea here is replacing express with angular itself for a server framework. Where you have a Server App that has a view which is the Universal app

gdi2290:
needs more research. What level of configuration should be needed?

* server level
* request level
* app level
* component level

jeffwhelpley:
This is being covered by #238

PR #73, opened 2015-08-09
module(server) : Create NodeBackend for Http #50

manekinekko:
module(server) : Create NodeBackend for Http #50 (WIP)


gdi2290:
here’s the part of the code where it’s parsing the url

parse url:
https://github.com/pwnall/node-xhr2/blob/master/src/001-xml_http_request.coffee#L636...L648

check if valid before sending:
https://github.com/pwnall/node-xhr2/blob/master/src/001-xml_http_request.coffee#L173...L188

that’s why we need a full url. It's to blame for not allowing relative urls

gdi2290:
related https://github.com/pwnall/node-xhr2/issues/18

manekinekko:
thx @gdi2290

manekinekko:
BTW, does the [http_server](https://github.com/angular/universal/blob/master/modules/server/src/http_server.ts) module have access to the preboot [options](https://github.com/angular/universal/blob/master/modules/preboot/src/interfaces/preboot_options.ts)? May be we could use these options to pass the URL?!

gdi2290:
app configuration should be set via (pending) `Environment` service https://github.com/angular/universal/issues/44 that way we can set the baseUrl allow with meta tags etc.

manekinekko:
Is the baseUrl considered to be the same as the one used when starting the Node server? I mean, is it the sever IP?

gdi2290:
node-xhr2 released a new version (v0.1.3) with baseUrl support

gdi2290:
you can remove all the trycatch calls. Can you also squash the commits into one

modules/server/src/helper.ts
```diff
@@ -52,3 +52,9 @@ export function selectorRegExpFactory(selector: string): RegExp {
 }
 
 
+export function relativeToAbsoluteUrl(relativeUrl: string): string {
+  var _server = require('../../../../index.js').Server;
+  var _addr = _server.address();
+  var _baseUrl = `http://${ _addr.address }:${ _addr.port }`;
```
gdi2290: this fine for now

gdi2290:
merged via https://github.com/angular/universal/commit/37a76589151f90b6a6264b7da2d3f4a45947f17c

PR #76, opened 2015-08-10
module(server) : Create NodeBackend for Http #50

manekinekko:
Task https://github.com/angular/universal/issues/50

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,5 @@
+var serverip = '127.0.0.1'
+
 var port = process.env.PORT    || 3000;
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -1,3 +1,5 @@
+var serverip = '127.0.0.1'
```
gdi2290: good for now

manekinekko:
Should we add support for a configuration file, such as `karma.config.js` or `jasmine.config.js` ? 
Let's call it `universal.config.js` or perhaps `preboot.config.js`. We can then give the IP address of the server.

PR #78, opened 2015-08-10
refactor(preboot) - add karma tests

### LINKED ISSUE #63: Move preboot tests back to modules/preboot/test
e2e tests are global and remain in root while module tests should remain within the module

jeffwhelpley:
This will close #63.


karma.conf.js
```diff
@@ -1,70 +0,0 @@
-// Karma configuration
-// Generated on Mon Jul 27 2015 11:34:50 GMT+0530 (IST)
-
-module.exports = function(config) {
```
gdi2290: do you think it's a good idea to remove this file? this doesn't allow people to run `karma` be itself. I understand the benefit of having one file for configuration but maybe we should export it from the gulpfile and import it here. What do you think?

jeffwhelpley:
My thinking is that we can add separate karma.conf later but I am not quite sure where else we would be using it besides preboot. All other code in this repo will likely be for the server side. So jasmine server side tests and protractor e2e tests should suffice. If we end up doing karma unit tests for some client side angular core feature in this repo then the karma config will be quite different. Also FYI I want to add code coverage back in a future change so I can move some common values that won't change to separate karma.conf then.

PR #81, opened 2015-08-11
fix(preboot): maintain cursor position

### LINKED ISSUE #68: focus on <input> cursor position differs between server rendered and client rendered
given an html element
```html
<input value="testing" autofocus [value]="value" (keyup)="value = $event.target.value">
```
the cursor position is at the beginning `[|testing]` for server rendered view. Then preboot.js transfers the focus back `[testing|]` the cursor position is now the latest position and value. Also note that angular doesn't save the cursor position when a user types in a new value. This probably due to setting the `[value]` on keyup

jeffwhelpley:
This fixes #68.

### DIFF
--- a/examples/preboot/preboot.html
+++ b/examples/preboot/preboot.html
@@ -7,12 +7,12 @@
   
   <body>
     <app>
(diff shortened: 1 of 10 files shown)


gdi2290:
LGTM, merging

Issue #82, opened 2015-08-12
e2e test for ServerDomRenderer and with preboot.js


gdi2290:
We need e2e tests for 

Server:
* default values for input value and checkbox
* focus an input element during Server rendering that adds autofocus

Server+Client:
* default values for input value and checkbox
* focus an input element during Server rendering that adds autofocus

Server+Client+Preboot.js
* default values for input value and checkbox
* focus an input element during Server rendering that adds autofocus
* selection is correct placement in an input
* selection is correct placement in an input during bootstrap

gdi2290:
I'm working on this now

gdi2290:
I'm not sure how to get element props with protractor to test client rendered elements and selection

manekinekko:
Hi @gdi2290 have you started this?

gdi2290:
@manekinekko I push up the wip. the problem is getting the props rather than the attrs
https://github.com/angular/universal/blob/master/test/test_page.e2e.js
you would need to inject js to grab elements and expose them to protractor I think

manekinekko:
@gdi2290 which props do you wanna test? I managed to get pros from DOM element.

gdi2290:
@manekinekko the input to ensure selection is correct and the value of the input

manekinekko:
@gdi2290 using `executeScript` you can run arbitrary js code. So for example, this gives you the `value` property:
```javascript
it('should have default value for inputs', function() {
  subject = browser.executeScript('return document.querySelector("#defaultValueInput").value');
  result  = 'value8';

  expect(subject).toBe(result);
});
```

gdi2290:
oh that's a good find and solves a lot of problems :+1:

manekinekko:
should I emphasis on the difference between the prop "value" and the attribute "value" in the test suites?

gdi2290:
yup, although for server only it's attrs while everything else it's props it's important to note that

manekinekko:
@gdi2290 here is what I was able to test so far. As you can see from the screenshot below, there is one test that fails and 3 pending test that I don't know how to test (the preboot ones). 
I let you check that if you dont mind ;)

If you're OK. I'll send the PR right away.

![image](https://cloud.githubusercontent.com/assets/1699357/9345570/b468350a-4612-11e5-8c42-58429b79039b.png)

gdi2290:
LGTM

gdi2290:
selection isn't really working with preboot or at least not with the e2e tests

[...]

Issue #25, opened 2015-07-17
Better ways to test Universal apps


gdi2290:
Consider different ways to construct e2e tests such as [Page Object Pattern](https://code.google.com/p/selenium/wiki/PageObjects) or [Bot Pattern](https://code.google.com/p/selenium/wiki/BotStyleTests) (conventions for protractor e2e)

gdi2290:
Consider a decoupling the view using other methods such as the one described in this post
https://medium.com/how-we-build-fedora/protractorjs-a-better-way-to-implement-page-objects-bc927cdb3f69

gdi2290:
similar to https://github.com/angular/universal/issues/44 where there is a test environment with helper syntax in html or js that should be removed in other envs

gdi2290:
This is for a future refactor:

Thinking about it more each test should be isolated for each `it` block. So each `it` block is testing a certain feature which means you can include a test for each state of the app.



something like 
```es6
state('should have checked values for checkbox inputs', {
  // should run anywhere
  universal: function() {
    // noop
    expect(subject).toEqual(result);
  },
  server: function() {
    // check attr
    expect(subject).toEqual(result);
  },
  client: function() {
    // check prop
    expect(subject).toEqual(result);
  },
  preboot: function() {
    // checked should transfer from attr to prop
    expect(subject).toEqual(result);
  }
}
```
Where you can noop or just not include a function for the env you're not testing for. Otherwise the code you're writing, for example, to check the `checked` value of an input is either in separate files or 3 different scroll positions in your large test file

gdi2290:
we can just close this for now since it's experimental

PR #96, opened 2015-08-18
test(test_page.e2e.js): e2e test for ServerDomRenderer and with preboot.js

manekinekko:
PR for task https://github.com/angular/universal/issues/82
There is one failing test and 3 pending ones that need to be checked.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -87,6 +87,10 @@ var paths = {
 
   specs: [
     'dist/**/*_spec.js'
+  ],
+
+  e2e: [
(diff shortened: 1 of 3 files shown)


gdi2290:
I'll check them out and merge it. Thanks! :+1:

gdi2290:
okay so the problem is that some tests work on all app states while others don't (client is prop only while server is attr and preboot fixes everything that isn't working from server to client). which is why we need a better solution https://github.com/angular/universal/issues/25 to manage our tests in all these different states

PR #97, opened 2015-08-19
fix(preboot): now checkboxes, radio buttons and select dropdowns work

### LINKED ISSUE #88: transfer input [checked] prop from server view to client view with preboot.js
when the view is rendered the checkbox keeps the checked state in the prop so if the user clicks on the checkbox before the client bootstraps then the state is not transferred. Same for unchecked

jeffwhelpley:
This will fix #88.


gdi2290:
merging

PR #104, opened 2015-08-30
feat(server): ServerFormDirective for NoJS support

gdi2290:
Needed for NoJS support

We don't have a way to inject this directive during bootstrap to improve the form otherwise we have to ask the user to include this next to their form directive bindings

### DIFF
--- a/modules/server/src/form_directives_server.ts
+++ b/modules/server/src/form_directives_server.ts
@@ -1,0 +1,33 @@
+import {Directive} from 'angular2/angular2';
+import {ElementRef} from 'angular2/angular2';
+import {Renderer} from 'angular2/render';
+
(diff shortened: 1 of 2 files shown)


modules/server/src/form_directives_server.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const APP_LOCATION: OpaqueToken = CONST_EXPR(new OpaqueToken('appLocation'));
+
+@Directive({
+  selector: 'form',
+  host: {
+    'method': 'POST'
+  }
+})
+export class ServerForm {
```
gdi2290: this needs to go in client folder and add the array `UNIVERSAL_DIRECTIVES`

Issue #119, opened 2015-10-21
include lodash dependency for preboot


gdi2290:

gdi2290:
done via https://github.com/angular/universal/commit/331626b3bc9cd37296f08c4d08c7c59c2d63ac60

PR #120, opened 2015-10-21
fix(preboot): Add lodash as preboot dependency

jeffwhelpley:


jeffwhelpley:
This will fix #119

modules/preboot/package.json
```diff
@@ -37,6 +37,7 @@
     "gulp-rename": "^1.2.2",
     "gulp-uglify": "^1.4.2",
     "q": "^1.4.1",
+    "underscore": "^1.8.3",
```
gdi2290: this should be `lodash` rather than `underscore` since you're importing it 
https://github.com/angular/universal/blob/master/modules/preboot/src/server/normalize.ts#L8

jeffwhelpley: ha. well, the APIs are similar and all the tests pass so I am not sure it matters. However, I will change to lodash.

tbosch: Hi,
mmh, preboot should be as small as possible. I bet underscore is bigger
than preboot itself.
Do we really need this dependency?

jeffwhelpley: the inline JavaScript that runs on the client is extremely small and doesn't contain any dependencies. Lodash and other dependencies in package.json are all used for the server side preboot component. The server side preboot component is used to generate the client side JavaScript. So, no worries about the size for these.

jeffwhelpley:
Sorry about the confusion here. Changed underscore to lodash. All should be good, but let me know if you see any issues.

tbosch:
Jeff, could you measure the overhead preboot gets in size by using lodash?

modules/preboot/package.json
tbosch: K

tbosch:
Never mind, just saw your other reply on the issue.

gdi2290:
merged via https://github.com/angular/universal/commit/331626b3bc9cd37296f08c4d08c7c59c2d63ac60

Issue #124, opened 2015-10-30
docs: Include component building guide


gdi2290:
There needs to be a list of best practices for universal component aka component that may run on the browser, server, or webworker.

jeffwhelpley:
Folding into #258

PR #138, opened 2015-11-06
docs(readme): best practices for components

gdi2290:
Best practices for components as bullet points

related https://github.com/angular/universal/issues/124

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,14 +14,15 @@
 
 # Table of Contents
 * [Modules](#modules)
-    * [Server](#server)
+    * [Universal](#universal)
     * [preboot.js](#prebootjs)
+* [Best Practices](#best-practices)
 * [What's in a name?](#whats-in-a-name)
 * [License](#license)
 
 # Modules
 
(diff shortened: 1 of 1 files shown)


zhaxi:
@gdi2290 could universal component be webworker component at same time? I see angular2 & universal export different bootstrap, could one codebase both support universal app and webworker app? I feel a little confused.

gdi2290:
So a Universal component can work on the Client/Server/Webworker hence the term universal. I can show an example of a Universal Angular 2 app that is server rendered then loaded on the Client within a WebWorker if you like.

zhaxi:
I'd like to see your example! And I have another related question, If I want to support android 4.4-(without webworker support) and android4.4+(with webworker support), It's seems to me that we need two sets of code, With Universal Component&other help, could it be possible that we only need one set of codes(just like view encapsulation in NG2 native&emulated) to ensure compatibiity&best performance, thanks!

zhaxi:
@gdi2290 when can I see your example?

PR #140, opened 2015-11-12
Update to angular-alpha-0.46

mikeallisonJS:
Update to alpha-0.46
Remove several module overrides and use angular default classes.
Clean up several compiler warnings.


mikeallisonJS:
yuck, conflicts. and it didn't look like the base had changed...

mikeallisonJS:
eh, I just realized I copy/pasta'd a preboot import that will kill this.

PR #152, opened 2015-12-16
feat: upgrade to ng2 beta

wesleycho:
- Upgrades dependencies to angular 2.0.0-beta.0 and rxjs 5.0.0-beta.0

Should note that there are 6 TS errors still that need to be looked at, but this should get most of the way to having Universal 2.0 beta ready.


wesleycho:
So there is one major typing issue that is causing 559 of the 561 errors in TS - Angular 2 providing the typings for `node/node.d.ts`, and we also have `node/node.d.ts` from tsd.json. Those two type files are conflicting.

jeffwhelpley:
@gdi2290 let me know what you think, but for the node typings issue I have manually modified the tsds in the past (i.e. to point to the angular 2 node type defs), but let me know what you think.

gdi2290:
I'm checking it out now

PR #163, opened 2016-01-06
fix(render): fix wrong import

tamascsaba:
Missing dependencies in preboot server eg.:
preboot/src/server/client_code_generator.ts (q, gulp-uglify etc.)

Dependencies are in preboot/node_modules


gdi2290:
thanks!

Issue #157, opened 2016-01-03
update examples


gdi2290:
The scripts being loaded on the client for the examples aren't working. We're also moving away from bower to npm

* [x] remove bower
* [x] update examples with node_modules

gdi2290:
done via https://github.com/angular/universal/commit/26f313358865c41c11421ba6097b0f975e2a8f58

PR #165, opened 2016-01-07
fix(bower): fix install location of bower_components

wesleycho:
- Specify location for install of bower_components

By default bower installs to `app/bower_components` - this change fixes the location to what appears to be the intended location.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory": "bower_components"
+}


gdi2290:
I'll merge this but we need to remove bower and use npm (see issue https://github.com/angular/universal/issues/157)

gdi2290:
screen shot 2016-01-06 at 9 53 46 pm
via https://github.com/angular/universal/commit/470b0ce425f0d9763f0fa2b72a46477a12285784

Issue #142, opened 2015-11-19
fix templateUrl using node http


gdi2290:
templateUrl uses XHR service

mcpDESIGNS:
@gdi2290 @jeffwhelpley Love what you guys are doing, great project/idea. I wasn't sure how to get in touch with you guys, but how could I join the team to start helping in contributing?

jeffwhelpley:
@mcpDESIGNS We would love some help! @gdi2290 and I are cleaning up the open issues this week, so we will have some specific stuff you can work on by the end of the week and will let you know. Thanks for reaching out!

mcpDESIGNS:
Great! Glad to help
I think making Angular 2 isomorphic is crucial, I'll contribute as much as I can !

Is there any documentation someplace I can read about how everything works & fits together so far?

jeffwhelpley:
The best intro right now is to watch the two talks we gave recently:

https://www.youtube.com/watch?v=0wvZ7gakqV4
https://www.youtube.com/watch?v=MtoHFDfi8FM

You can also read our origin design doc (though it is not 100% up to date):

https://docs.google.com/document/d/1q6g9UlmEZDXgrkY88AJZ6MUrUxcnwhBGS0EXbVlYicY/edit

mcpDESIGNS:
I was just reading through the docs, found you guys through the videos, great work!
Looks like I have plenty of reading to do then. Thanks and keep me in the loop!

gdi2290:
@mcpDESIGNS are you still interested in adding this feature? what you need to do is replace `XHRImpl` 
[modules/universal/server/src/platform/node.ts#L103...L104](https://github.com/angular/universal/blob/e7bdf86f04afd172bf4d9cb71db3764c13acacc3/modules/universal/server/src/platform/node.ts#L103...L104) with your own version `NodeImpl` that has the same interface as [modules/angular2/src/platform/browser/xhr_impl.ts](https://github.com/angular/angular/blob/89f32f808f061a5d70ecd4362c4f51d240014fa0/modules/angular2/src/platform/browser/xhr_impl.ts)

dbabaioff:
Fixed: https://github.com/angular/universal/pull/167

gdi2290:
<img width="62" alt="merged" src="https://cloud.githubusercontent.com/assets/1016365/12177951/c1b0628e-b523-11e5-8dfa-d1e19e6f19a8.png">
[...]

PR #167, opened 2016-01-07
fix(helper): backslash escaping (\s\S) in template strings are handled incorrectly

dbabaioff:
This fixes a issue with the selector with attrs RegExp and template string (https://github.com/angular/universal/commit/787d98c90dfdb713c1e80a97764c51424ade967b)

Before:
```js
var regExpSelect = "<" + escapeRegExp(selector) + "[^>]*>([sS]*?)</" + escapeRegExp(selector) + ">";
```

After:
```js
var regExpSelect = "<" + escapeRegExp(selector) + "[^>]*>([\\s\\S]*?)</" + escapeRegExp(selector) + ">";
```

See: https://github.com/Microsoft/TypeScript/issues/2803

### DIFF
--- a/modules/universal/server/src/platform/node.ts
+++ b/modules/universal/server/src/platform/node.ts
@@ -1,6 +1,9 @@
+import * as http from 'http';
+import * as url from 'url';
+
 // Facade
 import {Type, isPresent, CONST_EXPR} from 'angular2/src/facade/lang';
-import {Promise} from 'angular2/src/facade/promise'
+import {Promise, PromiseWrapper, PromiseCompleter} from 'angular2/src/facade/promise';
 
 // Compiler
 import {COMPILER_PROVIDERS, XHR} from 'angular2/compiler';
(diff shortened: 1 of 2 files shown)


dbabaioff:
(Committed a fix for: https://github.com/angular/universal/issues/142 on the same PR - wanted to open a separate PR)
Closes: https://github.com/angular/universal/issues/142

modules/universal/server/src/platform/node.ts
```diff
@@ -55,6 +58,41 @@ export function initNodeAdapter() {
...
 
+export class NodeXHRImpl extends XHR {
+  get(templateUrl: string): Promise<string> {
+    let completer: PromiseCompleter<string> = PromiseWrapper.completer(),
+      parsedUrl = url.parse(templateUrl);
+
+    http.get(templateUrl, (res) => {
+      res.setEncoding('utf8');
+
+      // normalize IE9 bug (http://bugs.jquery.com/ticket/1450)
```
gdi2290: lol we don't need to do this since we're in node.js environment but don't worry I'll merge it in

gdi2290:
LGTM, merging

modules/universal/server/src/platform/node.ts
dbabaioff: We still need to create a class for xhr (node.js), no?

gdi2290: I was talking about the IE9 comment

PR #198, opened 2016-01-16
Fixing `gulp build`

manekinekko:
filename in `tsconfig.json` was misspelled.

Renamed: 
`"modules/universal/server/src/platform/xhr_node_impl.ts",` 
to:
`"modules/universal/server/src/platform/node_xhr_impl.ts",`


gdi2290:
thanks!

PR #223, opened 2016-01-25
chore: remove bower

wesleycho:
- Remove all bower references

I don't see anymore reference to bower usage, so this strips out the vestiges.


gdi2290:
merged via https://github.com/angular/universal/commit/bbd35e27703845f531adc0c30524be79bbf725fa

Issue #86, opened 2015-08-13
Create Falcor/Data app example


gdi2290:
There should be a simple solution for working with "complex" models when dealing with server rendered views

jeffwhelpley:
@wesleycho completely this issue with 36f2dca

PR #224, opened 2016-01-25
feat(demo): add falcor demo

wesleycho:
- Add basic falcor demo

This is towards addressing #86 - this is still a WIP essentially, but safe as is.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -170,7 +170,7 @@ gulp.task('default', function(done) {
 
   var tasks = [
     'lint',
-    'karma'
+    'test'
   ];
 
   return $.runSequence(tasks, done);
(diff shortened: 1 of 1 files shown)


gdi2290:
can you use the same server?

wesleycho:
Yep - did a refactor

gdi2290:
We should mock out the api rather than including the lib but for now let's just get it in

wesleycho:
Wait...what github?

wesleycho:
Ah I see...confusing history

gdi2290:
merged via https://github.com/angular/universal/commit/36f2dcae6bdeaf3b2e77048f3126c839bf960f28

PR #229, opened 2016-01-27
fix(universal): published latest universal preview and fixed tsconfig

jeffwhelpley:
The universal starter needed a bump in version for universal preview. Also fixed tsconfig in universal for one of the test references.

@gdi2290 note that the universal module has some tsc warnings that we should fix in a subsequent fix.


gdi2290:
I'm using exclude now so the file is included but I'll release a new version

Issue #238, opened 2016-01-31
DOCUMENT provider refactoring


alexpods:
Currently provider for `DOCUMENT` requires `appComponentType` (see it [here](https://github.com/angular/universal/blob/master/modules/universal/server/src/platform/node.ts#L73)). It looks very strange to me, because we provide `appComponentType` only on bootstrapping phase, not when we're creating an application. Also this provider creates **mock** document which makes it very difficult to implement services for changing document title, meta tags and so on. I think it would be much better if we created a **real** document there based on a **real** html.

I've implemented this in `up` branch of [`angular2-universal-starter`](https://github.com/alexpods/angular2-universal-starter/tree/up).

**The main points:**
* Every `application` should be bound with a particular `DOCUMENT_HTML` (see it [here](https://github.com/alexpods/angular2-universal-starter/blob/up/src/.patches/universal/node.ts#L82) and [here](https://github.com/alexpods/angular2-universal-starter/blob/up/src/.patches/universal/render.ts#L34))
* Provider for `DOCUMENT` creates a **real** document based on this `DOCUMENT_HTML` (see it [here](https://github.com/alexpods/angular2-universal-starter/blob/up/src/.patches/universal/node.ts#L17))
* `parse5` serialization is used after bootstrapping (see it [here](https://github.com/alexpods/angular2-universal-starter/blob/up/src/.patches/universal/render.ts#L44)).
* [My benchmark](https://github.com/alexpods/angular2-universal-starter/blob/up/bench/renderComponent.ts#L38) shows the same speed as an original implementation.

**Pros**
* No need to provide `appComponentType` as an application provider.
* Now implementing services like `Title` becomes a trivial task. For example see `ServerTitle` implementation [here](https://github.com/alexpods/angular2-universal-starter/blob/up/src/.patches/universal/server_title.ts#L6). In this case if we have some `workers` page like:
```js
@Component({
  selector: 'worker',
  template: '<h1>Workers Page</h1>'
})
export class Workers {
  constructor(titleService: Title) {
    titleService.setTitle('Workers page!!!');
  }
}
```
then prerendered html will contain `<title>` tag with `"Workers Page!!"` text. And I want to mention that `<title>` tag is very important for search engines.

What do you think about it?

gdi2290:
We probably should add you to the repo since you understand what's going on. We need to support both ways of creating the document but your implementation is solid. We should probably megre your changes. There are a couple of things we need to sync up on

jeffwhelpley:
[...]

PR #243, opened 2016-02-02
feat(render): add renderDocument() and renderDocumentWithPreboot()

alexpods:
Relates to #238 

1. Because in [this comment](https://github.com/angular/universal/issues/238#issuecomment-177429694) @gdi2290 said that "we need to support both ways of creating the document" I've decided not to change `DOCUMENT` provider. Instead I've implemented `renderDocument()` and `renderDocumentWithPreboot()` functions.

2. `parseDocument` function should be implemented in core angular2 library [here](https://github.com/angular/angular/blob/master/modules/angular2/src/platform/server/parse5_adapter.ts#L557)

3. Parsing the same html on each request is not reasonable. But it can be easily optimized (and it should be).


gdi2290:
merged via https://github.com/angular/universal/commit/9b60c141ddd0e85fab29b17944f3607285850dd2

PR #270, opened 2016-02-14
fix(universal): get method should take parsed options instead of string

alfamegaxq:
When trying to use templateUrl in component, server throws exceptions. That's because http method receives path string instead of json options and therefore hostname in request is undefined.


gdi2290:
@alfamegaxq thanks, I'll cut a patch release

gdi2290:
@alfamegaxq published as `0.55.3`

PR #287, opened 2016-02-26
chore(all): Upgrade to Angular beta.7.

alxhub:


package.json
```diff
@@ -91,7 +91,7 @@
     "yargs": "^3.14.0"
   },
   "dependencies": {
-    "angular2": "2.0.0-beta.6",
+    "angular2": "2.0.0-beta.7",
```
gdi2290: you probably need to also update zone.js then we can do another PR for typescript

modules/angular2-gulp-prerender/package.json
```diff
@@ -32,7 +32,7 @@
     "through2": "^2.0.0"
   },
   "peerDependencies": {
-    "angular2": "2.0.0-beta.1 - 2.0.0-beta.6",
+    "angular2": "2.0.0-beta.7",
```
gdi2290: tbh we should only include `angular2-universal-preview` as the peer dep

gdi2290:
merged via https://github.com/angular/universal/commit/c4bbb1b6c9a2efa77ddba15dc9ab37e14754e05b

PR #301, opened 2016-03-04
feat: Multiple AppRoots for serverside rendering

gdi2290:
This code supports multiple appRoots for serverside rendering


gdi2290:
merged but accidently undid the commit

PR #345, opened 2016-04-03
Feat precache

gdi2290:
WIP precache


jeffwhelpley:
What's going on here with this PR? If we aren't going to merge it, let's kill it.

PR #368, opened 2016-04-18
refactor(ORIGIN_URL): Change to localhost

MarkPieszak:
With the relative api path fix recently, we can now hit
`/api/data.json` but this breaks because the server is on localhost &
the ORIGIN_URL is set to 127.0.0.1. Updated that fixes it


MarkPieszak:
@gdi2290 What's the easiest way to Squash these within a PR? Hate rebase...

gdi2290:
replace your fork with latest then rebasing from there would be easier

PR #381, opened 2016-05-03
chore: upgrade to angular 2.0.0-rc.0

wesleycho:
- Upgrade to Angular 2.0.0-rc.0

This upgrades universal to the latest Angular 2, but with a caveat - this PR is not complete due to the `@angular/animate` and `@angular/facade` packages not being published, or the code being available anywhere.

We cannot upgrade to the RCs until this is solved.


MarkPieszak:
Apparently the router has changed a little bit too? Wonder if that will affect things

wesleycho:
Hm, good catch - I could change it to the deprecated one, and someone else could look at changing it to the new router.

wesleycho:
@jeffbcross - if you're still investigating, I realized this is also incomplete in a couple of other areas, namely that the engines need changing of the script tags it generates for the HTML.

One example [here](https://github.com/angular/universal/blob/master/modules/express-engine/src/engine.ts#L149-L172).

jeffbcross:
@wesleycho thx I'll see if I can cover the rest

laskoviymishka:
@wesleycho Should it be referenced to rc1 version now?

wesleycho:
Yep, but Jeff is taking over looking at this

jeffwhelpley:
I am going to close this one because @jeffbcross and @gdi2290 where able to get over a couple challenges and get up to rc 0 yesterday. If Patrick and I can't get up to RC1 today or have issues with the new router, I will enter some new issues to track our progress there.

PR #424, opened 2016-05-31
chore(parse5): Update logic to work w/ parse5@2.*

MarkPieszak:


gdi2290:
@MarkPieszak what about the Parse5Adapter?  https://github.com/angular/angular/blob/2953ea10a7638d05b64486cdec5c6a240a8d675e/modules/%40angular/platform-server/src/parse5_adapter.ts

modules/universal/src/node/platform/dom/node_dom_renderer.ts
```diff
@@ -249,6 +251,16 @@ export class NodeDomRenderer extends DomRenderer {
     }
     return super.invokeElementMethod(location, methodName, args);
   }
+  
+  _setCheckedAttribute(renderElement, propertyName, propertyValue) {
```
gdi2290: can you make a separate PR for this so we can merge it sooner

MarkPieszak: Will do

PR #451, opened 2016-06-22
chore: replace private imports with copies of utility functions

jeffbcross:
This PR gets rid of many of the deep imports from Angular. For utility functions from Angular facade, I copied some of the useful ones into `universal/src/common/utilities.ts`. For simpler utility functions, I just inlined it.

This leaves just a few files that are using deep imports: 
 * modules/universal/src/node/platform/node_template_parser.ts
 * modules/universal/src/node/platform/node_template_parser-rc.0.ts
 * modules/universal/src/node/platform/node.ts

_Most_ of what those modules are importing from deep paths is types that are just used for type information.


gdi2290:
LGTM
`node_template_parser` is workaround to allow script tags for an index.html component. There might be a better way to do this now without replacing the whole service

Issue #478, opened 2016-07-13
Remove typings - use TS2 @types


MarkPieszak:
closes #477

PR #479, opened 2016-07-13
chore(typings): Remove typings - Add @types

MarkPieszak:
Moved typed-xhr2 to custom_typings folder to remove typings completely.

Upgrade to Typescript@beta (2)

Closes #479 #478


package.json
```diff
@@ -99,9 +100,16 @@
     "rxjs": "5.0.0-beta.6",
     "through2": "^2.0.0",
     "ts-node": "^0.9.3",
-    "typescript": "^2.0.0-dev.20160706",
-    "typings": "^1.0.5",
+    "typescript": "^2.0.0",
     "xhr2": "^0.1.3",
     "zone.js": "~0.6.12"
+  },
+  "dependencies": {
```
gdi2290: can you keep these in devDeps

tsconfig.json
```diff
@@ -23,10 +23,21 @@
...
+      "core-js",
+      "es6-promise",
+      "jasmine",
+      "node",
+      "through2",
+      "zone.js"
+    ]
   },
 
   "files": [
```
gdi2290: there should be glob support here now. can you check

types.d.ts
```diff
@@ -0,0 +1,7 @@
+/// <reference path="node_modules/@types/browserify/index.d.ts" />
```
MarkPieszak: I'm not sure, was talking to Jeff about this he said VSCode has some issues with these things. I have the latest 1.3.1. I can remove this file if you don't think we need it then. @gdi2290

tsconfig.json
MarkPieszak: I don't think these types are necessary they just help the compiler? Should I remove them for now?
Not sure on globs impossible to find any good info on types at the moment.

Just saw a post from @DanielRosenwasser ([here](https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/#comment-19085))
The compiler will automatically find global declarations from the first @types folder, and if you you need the compiler to search a little farther, there will be a ‘types’ property in ‘package.json’ to specify that you need a specific set of files, which would look something like this:
```
{
"compilerOptions": {
"types": ["jasmine", "jquery"]
}
```
Think he meant tsconfig, (types option shows up there)
image

package.json
MarkPieszak: Done

types.d.ts
gdi2290: let's keep it

package.json
```diff
@@ -70,6 +70,13 @@
     "@angular/platform-server": "2.0.0-rc.4",
     "@angular/router": "3.0.0-alpha.8",
     "@angular/router-deprecated": "2.0.0-rc.2",
+    "@types/jasmine": "^2.2.28-alpha",
+    "@types/browserify": "^12.0.28-alpha",
+    "@types/core-js": "^0.9.27-alpha",
+    "@types/es6-promise": "0.0.27-alpha",
```
gdi2290: we have core-js so we can remove es6-promise

gdi2290: if we're not including core-js anywhere we can stick with es6-promise or use the `"lib"` prop in tsconfig.

gdi2290:
can you also create index.d.ts with
```
declare module '*';
```

gdi2290:
@MarkPieszak add declare module '*'; then it LGTM

MarkPieszak:
Is it okay that it throws a ts error `'{' is expected` ?

gdi2290:
LGTM if `declare module '*';` works

PR #498, opened 2016-07-30
bug(NodeDomRenderer): Fix false flag for checked

MarkPieszak:
- [X] universal

Was `else if (propertyValue = false)` needs the triple `===`


gdi2290:
ok this makes sense

PR #546, opened 2016-09-19
Express engine config - add preboot options

MarkPieszak:


gpazo:
any ETA when this will be pushed to npm? or a workaround for now?

MarkPieszak:
@gpazo This is related to #552 which could handle actual preboot, this is just to fix the Interface object. Preboot was missing as one of the objects there.

Useage:

```
// Our Universal - express configuration object
  const expressConfig : ExpressEngineConfig = {
    req,
    res,
    ngModule: MainModule,
    preboot: false,
    baseUrl: '/',
    requestUrl: req.originalUrl,
    originUrl: 'http://localhost:3000'
  };

  // NOTE: everything passed in here will be set as properties to the top level Zone
  // access these values in your code like this: Zone.current.get('req');
  // this is temporary; we will have a non-Zone way of getting these soon
  res.render('index', expressConfig);
```

gpazo:
thanks @MarkPieszak ! I had a typo in my config object, the example you posted worked for me. 👍

Issue #550, opened 2016-09-21
NodePlatform's serializeModule returns a promise that resolves even if there was an error


SteveSandersonMS:
* **I'm submitting a ...**
It's either a bug report or a feature request, depending on what the intended design is.

* **What modules are related to this Issue?**
- [x] universal

* **What is the current behavior?**

Given the following:

```
import { platformNodeDynamic } from 'angular2-universal';
const prerenderPromise = platformNodeDynamic().serializeModule(...);
```

... the `prerenderPromise` always resolves successfully, even if there was an error in the Angular app while serializing the module (e.g., a component's constructor threw, or there was a syntax error in a template that was being rendered).

* **What is the expected behavior?**

I'd expect the promise to be rejected if there was an error.

* **What is the motivation / use case for changing the behavior?**

Robust error handling.

* **Please tell us about your environment:**

All

* **Other information**

I'm building the ASP.NET Core SPA templates (which @MarkPieszak is very familiar with). Our server-side boot logic used to look like [this](https://github.com/aspnet/JavaScriptServices/blob/8f550c5706158dfb36a99c713b377acedbba8c47/templates/Angular2Spa/ClientApp/boot-server.ts), but I noticed that if you introduce an error into the app (e.g., try to evaluate `someNonExistentFunction()` in a template), then it just logs the error to the console but resolves the promise as successful, so the app renders without any obvious feedback that anything is wrong. Of course this makes it tricky for developers to understand any problems they might be experiencing.

To fix this, I've made the following change to capture Zone errors and force the promise to reject: https://github.com/aspnet/JavaScriptServices/commit/5214a553a7d98bc532b956af7ad14b8905302878?w=1

Could you provide feedback on:

[1] Is it really intended for `serializeModule` not to reject its promise on error? If so, to make it easier for developers to understand, could we have a flag called something like `abortOnError` in the `config` (the object that includes `baseUrl`, `preboot`, `document`, etc.) that configures it to reject the promise if there's an error in the zone?

[2] Possibly better still, have you considered exposing an API that encapsulates the plumbing around `Zone` etc. to simplify things further? For example, it could look like:

```
const prerenderPromise = platform.serializeModuleInZone(AppModule, {
    baseUrl: '/',
    document: '...',
    ...
});
```

Internally it would just call `Zone.current.fork` with an `onHandleError` that rejects the promise. Then developers don't even need to see the obscure-looking Zone stuff.

[...]

PR #551, opened 2016-09-21
bug(platform-node): Throw error to parent catch of serializeModule

MarkPieszak:
Helps pass errors to `.catch()` of `Zone.run( ... serializeModule() )`

#550


gdi2290:
did you test that this works?

gdi2290:
I would think the error catching have to be in asyncPromiseSeries

MarkPieszak:
It seemed to work with random errors I tried throwing but should look into it more. In a wedding today so can't but could try it out more tomorrow! 

Isn't there already a catcher in the async series that would bubble up?

gdi2290:
@MarkPieszak any updates on this?

MarkPieszak:
@gdi2290 We should be all set (for now) with this at least from this end. It passes any errors thrown up to the caller.

gdi2290:
LGTM

gdi2290:
this is needed for asp.net

PR #552, opened 2016-09-21
feat(universal): Allow for configuring if preboot complete() is called automatically (immediatelly) or manually (asynchronously)

steve8708:

- [ ] express-engine
- [ ] grunt-prerender
- [ ] gulp-prerender
- [ ] hapi-engine
- [ ] universal-next
- [x] universal
- [ ] webpack-prerender

Feature

preboot complete() is called immediately upon client bootstrap, wether you want it to be or not

Allows a configuration to override the automatic prebooting. This is especially useful is you have to do anything asynchronous before preboot complete runs, such as fetching data.

Usage is
```ts
imports: [
  UniversalModule.withConfig({ autoPreboot: false }), // default behavior is still `true`
]
```

No

* **Other information**:

Open to any feedback on other ways to implement this feature and if good where to add tests and docs. Thanks!

…

### DIFF
--- a/modules/universal/src/browser/universal-module.ts
+++ b/modules/universal/src/browser/universal-module.ts
@@ -42,6 +42,7 @@ export class OpaqueToken {
 const SharedStylesHost: any = __platform_browser_private__.SharedStylesHost;
 
 export const UNIVERSAL_CACHE = new OpaqueToken('UNIVERSAL_CACHE');
+export const AUTO_PREBOOT = new OpaqueToken('AUTO_PREBOOT');
 
 @NgModule({
   imports: [
@@ -77,15 +78,22 @@ export const UNIVERSAL_CACHE = new OpaqueToken('UNIVERSAL_CACHE');
       deps: []
     },
     {
+      provide: AUTO_PREBOOT,
+      useValue: true
+    },
+    {
       multi: true,
       provide: APP_BOOTSTRAP_LISTENER,
-      useValue: () => {
-        let _win: any = window;
-        if (_win && prebootClient) {
-          setTimeout(() => prebootClient().complete());
-        }
-      }
-    }
+      useFactory: (autoPreboot: boolean) => {
+        return () => {
+          let _win: any = window;
+          if (_win && prebootClient && autoPreboot) {
+            setTimeout(() => prebootClient().complete());
+          }
+        };
+      },
+      deps: [ AUTO_PREBOOT ],
+    },
   ]
 })
 export class UniversalModule {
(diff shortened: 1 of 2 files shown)


steve8708:
CLA signed 👍

jeffwhelpley:
@steve8708 sorry, can you change this so that it auto completes by default? Right now if you don't set autoPreboot then it won't complete. This will break everyone.

MarkPieszak:
Also erroring here in Travis `tsc`:

```
export function createGlobalProxy() {
                    ~~~~~~~~~~~~~~~~~
modules/universal/src/node/proxy-document.ts(116,17): error TS7030: Not all code paths return a value.
```

jeffwhelpley:
@ca136 not if you don't set the AUTO_PREBOOT token. You are introducing that with this change, so when someone else upgrades, they won't have that set and will thus not have autoPreboot set even though they technically did have it before. Typically we want to try and ensure that when changes are introduced someone can upgrade without adversely affecting behavior. If you upgraded to this but didn't change anything else, preboot would go from automatically completing to not automatically completing.

gdi2290:
LGTM

steve8708:
thanks @gdi2290 👍 

@MarkPieszak looks like that tsc error was already on development but added another commit to this PR that resolves the error 👌

modules/universal/src/browser/universal-module.ts
```diff
@@ -77,15 +78,22 @@ export const UNIVERSAL_CACHE = new OpaqueToken('UNIVERSAL_CACHE');
       deps: []
     },
     {
+      provide: AUTO_PREBOOT,
+      useValue: true
```
steve8708: yeah, including the token in the default providers array here allows the default behavior to remain unchanged. 

anyone who doesn't provide a value for this token will get the default value `true` which will keep the current behavior the same for current universal users

only if someone explicitly provides a value for this token or provides a value for autoPreboot in the config object of UniversalModule.withConfig does this get overridden

andresantos123:
@steve8708 with this PR I disabled the preboot, but how it can be called now manually? The import of prebootClient from the old versions dont work anymore.

steve8708:
@andresantos123 come to think of it it may be nice to optionally export prebootComplete again like we did before

In the meantime you can see how we get the preboot client in universal [here](https://github.com/angular/universal/blob/6b71fd94da9aa7650c87f88400fca0606fa56a8f/modules/universal/src/browser/universal-module.ts#L22) (A little ugly unfortunately, I believe it is this way to support a breaking API change among different versions of the preboot library) and we call it [here](https://github.com/angular/universal/blob/6b71fd94da9aa7650c87f88400fca0606fa56a8f/modules/universal/src/browser/universal-module.ts#L63);

So you should be able to do similar

```ts
var prebootClient;
try {
  prebootClient = require('preboot/__build/src/browser/preboot_browser');
  prebootClient = (prebootClient && prebootClient.prebootClient) || prebootClient;
} catch (e) {}

// .. and when you are ready
prebootClient().complete()
```

PR #557, opened 2016-09-27
fix(platform-node): Remove memory leak

steve8708:

- [ ] express-engine
- [ ] grunt-prerender
- [ ] gulp-prerender
- [ ] hapi-engine
- [ ] universal-next
- [x] universal
- [ ] webpack-prerender

Fixes a memory leak

If you run servers in production, NgModules are generated per request and cached indefinitely, leading to a serious memory leak that will eventually floor your servers and you'll have to restart them constantly (up to every 15 minutes!)

Use WeakMap instead of Map to ensure if all references are lost to a cache key the value is garbage collected

No

* **Other information**:

### DIFF
--- a/modules/platform-node/node-platform.ts
+++ b/modules/platform-node/node-platform.ts
@@ -118,7 +118,7 @@ function s4() {
               /*implements PlatformRef*/
 export class NodePlatform  {
   static _noop = () => {};
-  static _cache = new Map<any, any>();
+  static _cache = new WeakMap<any, any>();
   get platformRef() {
     return this._platformRef;
   }
@@ -332,7 +332,7 @@ export class NodePlatform  {
         let prebootCode = null;
         // TODO(gdi2290): hide cache in (ngPreboot|UniversalPreboot)
         let prebootConfig = null;
-        let key = (typeof preboot === 'object') && JSON.stringify(preboot) || null;
+        let key = (typeof preboot === 'object') && preboot || null;
         let prebootEl = null;
         let el = null;
         let lastRef = null;


gdi2290:
LGTM, thanks!

steve8708:
Rad, thank you!

modules/platform-node/node-platform.ts
```diff
@@ -332,7 +332,7 @@ export class NodePlatform  {
         let prebootCode = null;
         // TODO(gdi2290): hide cache in (ngPreboot|UniversalPreboot)
         let prebootConfig = null;
-        let key = (typeof preboot === 'object') && JSON.stringify(preboot) || null;
+        let key = (typeof preboot === 'object') && preboot || null;
```
gdi2290: so the goal of this cache was reusing the same prebootEl but I guess the biggest problem was probably not detaching it from the dom after destroying the app. If we move preboot to `createEngine()` which is the plan then this is fine since we're maintaining the same reference to the preboot config. During each req/res we're creating a new preboot config which was why I wanted to ensure if the configs are the same that we reuse the preboot element

steve8708: ah, I was curious if that was why this was being stringified! Makes sense, and I frankly would've left it but WeakMaps don't support non-object keys so to get the memory releasing map it has to be the original object

Until moving to `createEngine()`, could also consider using an LRU or LFU cache in that case if you want to have serialized keys in this cache without risks of memory leaks. happy to send a followup PR if there is any other way you want this implemented here 👍

PR #560, opened 2016-09-28
feat(universal): AOT support

steve8708:

- [ ] express-engine
- [ ] grunt-prerender
- [ ] gulp-prerender
- [ ] hapi-engine
- [ ] universal-next
- [x] universal
- [ ] webpack-prerender

Add AOT support for the UniversalModule

Compiling with `tsc` doesn't generate the necessary `*.metadata.json` files for AOT support. Additionally some code updates were required for `ngc` to properly transpile the universal source

Use `ngc` instead of `tsc` to transpile typescript.

No

* **Other information**:


steve8708:
Remaining work to do

* [ ] Remove `src/` from transpiled resulting code in dist
* [ ] Add back caching for gulp watch (there is no gulp-typescript equivalent of `ngc`, at this time just seems to be a command line utility, which is unfortunate as that allows no caching for things like `gulp watch`, but there may be a workaround by using some of the `@angular/compiler-cli` internal code similar to the implementation by @mgechev [here](https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/seed/compile.ahead.prod.ts). Alternatively `ngc` could be used just for production builds and development could still use `gulp-typescript`. I'll look into both options and open to feedback for this as well
* [ ] Code cleanup

MarkPieszak:
Does this line work here for running it from gulp?
http://stackoverflow.com/questions/39340283/angular-2-ahead-of-time-compiler-with-gulp-typescript

var exec = require('child_process').exec;

```
gulp.task('task', function (cb) {
  exec('ngc -p "<path to your tsconfig.json>"', function (err, stdout, stderr) {
    console.log(stdout);
    console.log(stderr);
    cb(err);
  });
});
```

steve8708:
@MarkPieszak yeah that's exactly what I do in this PR, which totally works, just is not as efficient as `gulp-typescript` when in dev/watch mode since it transpiles your whole project on every change instead of just the changed file like universal did before

definitely not the end of the world, and if we're ok with that caveat then this PR should be good to go once I resolve the `src/` directory change (and address any other feedback)

gulpfile.ts
```diff
@@ -51,40 +49,22 @@ gulp.task('_test', () => {
...
-      .on('end', maybeDone);
-
-    function maybeDone() {
-      doneCount++;
-      if (doneCount === 2) {
-        resolve();
+    // TODO: better watch support to not transpile the whole project
+    // for single file changes, perhaps using similar implementation to
+    // this: https://github.com/mgechev/angular2-seed/blob/master/tools/tasks/seed/compile.ahead.prod.ts
+    child_process.exec('node_modules/.bin/ngc -p tsconfig.aot.json', (err) => {
```
steve8708: @MarkPieszak this is the line relevant to your comment regarding running ngc with gulp

gulpfile.ts
```diff
@@ -51,40 +49,30 @@ gulp.task('_test', () => {
...
   return new Promise((resolve) => {
-    var doneCount = 0;
-    output.js
-      .pipe(rename(buildUtils.stripSrcFromPath))
-      .pipe(replace(/\.\/src\//g, './'))
-      .pipe(gulp.dest('dist'))
-      .on('end', maybeDone),
-    output.dts
-      .pipe(rename(buildUtils.stripSrcFromPath))
-      .pipe(replace(/\.\/src\//g, './'))
```
gdi2290: we need to rename files still and remove the src folder

felipedrumond:
After migrating from RC4 to release, AoT support for Universal is what I'm waiting in order to drop the bundle sizes of my application by performing effective tree-shaking.

gulpfile.ts
steve8708: definitely will update 👍 

just working through a couple more issues when using ngc to compile an app depending on the universal compiled code and will get on getting the `src/` directly back out of the dist code 👌

gdi2290:
@steve8708 do you think we can pair on this sometime in the next few days? If you're busy I can also take over the PR

steve8708:
@gdi2290 happy to

I've currently got @qdouble's boilerplate running with universal AOT locally, so as of right now, everything is tested and working in this PR and the only update really blocking this PR (besides any review updates) is getting the src/ folder removed. 

I can add a quick and dirty script to remove the src folder and regex files to not refer to it, unless there is a more elegant way you have in mind (gulp certainly would be great if ngc supported gulp streams, or perhaps using a webpack build with `@ngtools/webpack` would be a more robust option worth considering). Happy to hear your input on that or let you run with it as you prefer

Definitely don't hesitate to pull this down and make any updates you like. Some of my findings that may be useful -
* sometimes ngc compiles but creates code that won't work (e.g. the @internal issue is only noticed when linking angular2-universal and running ngc in a parent repository that depends on universal, ngc will not be able to "find" anything marked `@internal` and won't compile, even if it's only within the angular2-universal source that ever references code tagged with `@internal` in comments). I've been using [this branch](https://github.com/qdouble/angular2webpack2-starter/compare/master...steve8708:aot-dirty?expand=1) to test out ngc from a parent repository while npm linkedin, as well as another one in a private repo that is much larger to test scenarios like this out
* ngc refuses to properly when using __internal_imports__, saying it can't find the necessary metadata. This doesn't seem right, but the only workaround that has worked this far is importing from file paths directly though I assume you are trying not to do this anymore given they were all replaced in by __internal_imports__ in you rc.5+ support PR(s)
* when importing from universal in parent repo's with aot you need to be sure to specify `import { /* ... */ } from 'angular2-unviersal/node' and `import { /* ... */ } from 'angular2-unviersal/browser' instead of just 'angular-universal'

Also feel free to ping me in the universal slack room (I'm @steve)

steve8708:
@gdi2290 this is now updated to properly remove the `src/` folder with gulp streams similar to how it was done before

I tried moving back to `__private_imports__` exactly as angular does internally but ngc refuses to compile saying that `[module] is not a StaticSymbol` just like in [this open issue](https://github.com/angular/angular/issues/11276). 

I'm created a reproducible case in that thread so the issue can be resolved with the compiler cli and we should be able to move back to using `__private_imports__` once it is

steve8708:
Actually, one last issue I've uncovered when hooking this up to a larger project

One small but odd interface incompatibility issue with a universal service that replaces a browser one in the DI when compiling with ngc. Aiming to have the issue resolved and this PR 100% ready to go sometime tomorrow

gdi2290:
kool, thanks looking forward to it

MarkPieszak:
@gdi2290 won't this be an issue? 
Wonder why it won't work with the private imports

```
+// Private imports
 +// Ngc does not compile properly when taking these from __private_imports__ but works
 +// as expected when importing from their full file path
 +import { KeyEventsPlugin } from '@angular/platform-browser/src/dom/events/key_events';
 +import { DomEventsPlugin } from '@angular/platform-browser/src/dom/events/dom_events';
 +import { HammerGesturesPlugin } from '@angular/platform-browser/src/dom/events/hammer_gestures';
 +import { DomSanitizerImpl } from '@angular/platform-browser/src/security/dom_sanitization_service';
 +import { DomRootRenderer } from '@angular/platform-browser/src/dom/dom_renderer';
 +import { SharedStylesHost, DomSharedStylesHost } from '@angular/platform-browser/src/dom/shared_styles_host';
```

tsconfig.aot.json
```diff
@@ -0,0 +1,74 @@
...
+    "removeComments": true,
+    "baseUrl": ".",
+    "emitDecoratorMetadata": true,
+    "experimentalDecorators": true,
+  	"sourceMap": true,
+    "inlineSources": true,
+    "rootDir": "./modules",
+  	"outDir": "compiled/ngc",
+    "pretty": true,
+    "stripInternal": true,
```
alexeagle: rather than remove all the `@internal` can you just set this false?

steve8708: @alexeagle this is spot on!

I didn't even notice this option, but setting it to false has resolved both the `@internal` and `__private_imports__` issues that were driving me crazy

thank you for the tip!

tsconfig.aot.json
```diff
@@ -0,0 +1,74 @@
+{
```
alexeagle: copy-pasta of this whole file is going to be a maintenance hassle. Why not just update tsconfig.json as needed?

steve8708: originally was because it was a convention I was seeing among other projects using aot, and later was because the files array and rootDir option were differnet for the aot config vs the general tsconfig (e.g. setting the rootDir to the "modules' directory and skipping other ts files like gulpfile.ts, etc)

I can update this by adding another pipe to the gulp stream though to remove the modules directory and reset the rootDir back to the way it was

will see if there are any other complications and if not I'll add another commit getting this back into the same file 👍

modules/express-engine/src/index.ts
```diff
@@ -1,6 +1,10 @@
 const fs = require('graceful-fs');
 
-import { platformUniversalDynamic } from 'angular2-universal';
+// TODO: TS can't find module 'angular2-universal/node' when using direct imports
```
alexeagle: If you use `paths`
```
      "angular2-universal/node": [
        "./modules/universal/src/node"
      ],
```

then you can go back to the import, and we still get type-checking

tsconfig.aot.json
alexeagle: I created the aot configuration (as a sibling block of `"compilerOptions"`) this way specifically to avoid proliferating the configuration...

I don't see why the rootDir, outDir, or files need to be different. Can't we compile the same way as before, but just call `ngc` rather than `tsc`?

modules/express-engine/src/index.ts
steve8708: Absolutely, but this unfortunately hasn't been behaving as expected. We already have that in our paths but ngc just doesn't seem to recognize it

In compiler of tsconfig.json (and tsconfig.aot.json) we have

```json
    "paths": {
      "angular2-express-engine": [
        "./modules/express-engine"
      ],
      "angular2-platform-node": [
        "./modules/platform-node"
      ],
      "angular2-universal/browser": [
        "./modules/universal/browser.ts"
      ],
      "angular2-universal/node": [
        "./modules/universal/node.ts"
      ],
      "angular2-universal": [
        "./modules/universal"
      ],
      "angular2-hapi-engine": [
        "./modules/hapi-engine/src/index.ts"
      ]
    }
```

but switching back to import from require always emits

```
[09:48:58] Error: Command failed: node_modules/.bin/ngc
Error: Error at /Users/steve/Projects/forks/universal/modules/express-engine/src/index.ts:3:42: Cannot find module 'angular2-universal/node'.
```

Unless perhaps I'm missing something obvious?

alexeagle:
replacing "./modules/universal/node.ts" with "./modules/universal/src/node"
is what I tried, and it fixes the type-check error with tsc

steve8708:
@alexeagle perfect, thanks!

steve8708:
worked perfectly 👍

tsconfig.aot.json
steve8708: in this case it's because transpiling before was done with `gulp-typescript` and the `rootDir` was set outside of the tsconfig.json file to be `modules` and that plugin didn't complain about having files outside of that directory in our `files` list (probably does some dynamic setting of typescript settings under the hood when you set the rootDir through that plugin, I'm guessing)

run tsc with rootDir: "modules" in tsconfig complains that you can't have files like gulpfile.ts outside of that directory, and excluding gulpfile.ts, etc from the tsconfig file removes it from IDEs like vscode from inheriting the correct type checking settings that we want

just pushed a commit that works around this problem by allowing ngc to transpile everything to a temporary directory, then pulling in just the "modules" folder of the ngc output and doing some post processing with gulp before writing the files to our dist directory and all works as expected 👌

steve8708:
@alexeagle somewhat yes but not really

we used to use `gulp-typescript` which was really nice as we could use gulp and caching, rather than recompile everything on every change including multiple read/write streams on the actual file system

but because ngc is CLI-only right now, it breaks those niceties and regardless we need to transpile the whole project to an intermediate directory and then have gulp read from that and stream like we usually do with processing to it's final destination in dist

in an ideal world there would be an equivalent (or compatible) ngc plugin like (or using) gulp-typescript and then our dev flow would be much nicer (and faster, esp. when file watching)

but in the meantime, this does work just fine and I'm hoping more tools than just the `ngc` CLI will be coming out shortly (e.g. I'm really exciting to see the `@ngtools/webpack` stuff Rob, etc have been working on)

steve8708:
Last issue still remains of interface incompatibilities that are unexpectedly difficult to resolve in the resulting ngfactory code. Pinged @alexeagle on slack to help look into it and made a workflow to reproduce the issue

Hopefully we'll get this resolved today and can merge this in!

steve8708:
Everything should now be ready and working, no major blockers on my end besides getting a test fixture in here to ensure no new code is added that is broken post-ngc-compile like the main angular repo does

PR #531, opened 2016-09-15
chore(test): add initial unit tests

devCrossNet:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/angular/universal/blob/master/CONTRIBUTING.md#commit-message-format
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

* **What modules are related to this pull-request**
- [ ] express-engine
- [ ] grunt-prerender
- [ ] gulp-prerender
- [ ] hapi-engine
- [ ] universal-next
- [x] universal
- [ ] webpack-prerender

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
adding possibility to run jasmine tests and some unit tests for platform-node

* **What is the current behavior?** (You can also link to an open issue here)
no tests

* **What is the new behavior (if this is a feature change)?**
some unit tests

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
No

* **Other information**:

PR #567, opened 2016-09-30
fix(NodeDomRenderer): ensure attributes are string value

gdi2290:


gdi2290:
@jeffbcross can we merge this for now tests are done in another PR
https://github.com/angular/universal/pull/531

PR #571, opened 2016-10-03
chore(examples/webpack): config

gdi2290:


scott-wyatt:
It would be super useful to also have a working example with webpack-dev-server, I can't seem to get it working.

Issue #589, opened 2016-10-12
NodeConnection always set User-Agent to Angular 2 Universal


AlexYankee:
**Note: for support questions, please use one of these channels:** https://github.com/angular/universal/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports. Also, **Preboot has moved** to https://github.com/angular/preboot - please make preboot-related issues there.

* **I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support request here, see note at the top of this template.

* **What modules are related to this Issue?**
- [ ] express-engine
- [ ] grunt-prerender
- [ ] gulp-prerender
- [ ] hapi-engine
- [ ] universal-next
- [x] universal
- [ ] webpack-prerender

* **Do you want to request a *feature* or report a *bug*?**
I'm reporting a bug in NodeConnection 

* **What is the current behavior?**
If no UserAgent header present, NodeConnection constructor populates it with "Angular 2 Universal". It assumes header name to be equal "User-Agent". In fact, Angular2 normalizes headers names to be lowercase - 'user-agent'. That makes UserAgent header to always be set to Angular 2 Universal, no matter was header provided or not. For User-Agent sensitive APIs (blocking bots etc) that can lead to the wrong behavior.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** by creating a github repo.

* **What is the expected behavior?**
NodeConnection doesn't override User-Agent header if provided

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**

- Angular version: 2.0.1
- Browser: all
- Language: TypeScript
- OS:  Mac OS X
- Platform: NodeJs

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

kuldeepkeshwar:
@AlexYankee you can check #525

gdi2290:
thanks, fix is in a PR here https://github.com/angular/universal/pull/594

PR #594, opened 2016-10-12
fix(node-http): correctly set user-agent with normalized value

gdi2290:
closes https://github.com/angular/universal/issues/589


gdi2290:
@MarkPieszak thanks

PR #596, opened 2016-10-17
fix(node-platform): provide ResourceLoader for Templates that aren't inlined

jkuri:
- [ ] add simple test for that fix

This adds fix that templates does not have to be inlined.
This PR still need discussion if it's available to use this without `moduleId: module.id` option to get relative paths to templates. After chatting with Patrick there might be an option to let AoT compiler inline all templates.

\cc @gdi2290

### DIFF
--- a/modules/platform-node/node-platform.ts
+++ b/modules/platform-node/node-platform.ts
@@ -42,11 +42,12 @@ import {
   NgZone,
   Compiler,
   CompilerFactory,
-  TestabilityRegistry
+  TestabilityRegistry,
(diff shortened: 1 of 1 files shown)


gdi2290:
AoT should be inlining the templates if it's not already. This is to solve for the case where you might be using SystemJS or something without inlining these resources

gdi2290:
can you update the branch to latest again.

jkuri:
@gdi2290 rebased.

demurgos:
Could you publish the version with the fix to npm ?

PR #610, opened 2016-10-25
fix(gulpfile): ngc cross-platform fix

### LINKED ISSUE #608: 'ngc' is not recognized as an internal or external command,
- **I'm submitting a ...**
- [X] bug report
- OS:  **Windows** (Note: works fine on MAC etc)

On Windows, the `gulp` command can't properly run and breaks on Build during the spawn childprocess exec of **`ngc`**.

Looking into alternatives to get it working cross-platform.

Worth looking into alternative ways of running _ngc_ such as the way @mgechev does it [here](https://github.com/mgechev/angular-seed/blob/faa9f1929c3ea4c16ebc60e1dad47913acb7db83/tools/tasks/seed/compile.ahead.prod.ts)
@steve8708

MarkPieszak:
Fixes #608

'ngc' is an unknown internal/external command. 

Cross-platform fix to get
Windows `gulp` working by running ngc command through npm scripts.

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -58,7 +58,7 @@ gulp.task('build:no-clean', build);
 
 function build() {
   return new Promise((resolve, reject) => {
-    child_process.exec('node_modules/.bin/ngc', (err, stdout, stderr) => {
+    child_process.exec('npm run ngc', (err, stdout, stderr) => {
       if (err) {
         reject(err);
         return;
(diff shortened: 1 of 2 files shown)


gulpfile.ts
```diff
@@ -58,7 +58,7 @@ gulp.task('build:no-clean', build);
 
 function build() {
   return new Promise((resolve, reject) => {
-    child_process.exec('node_modules/.bin/ngc', (err, stdout, stderr) => {
+    child_process.exec('npm run ngc', (err, stdout, stderr) => {
```
steve8708: you may want to modify line 47 of this file as well as it should have the same bug in windows due to the path format being in unix style

gdi2290:
thanks!

PR #614, opened 2016-10-31
fix(node-document): create conditional for non html component

### LINKED ISSUE #347: Support rendering just a single component rather than a complete HTML document
As discussed, please add support for rendering just a single component element to an HTML string, instead of requiring the template to be an entire HTML document.

For example, instead of this:

```
ngUniversal.bootloader({
    ...,
    template: '<!DOCTYPE html>\n<html><head></head><body><app></app></body></html>'
})
```

... it should be possible to do this:

```
ngUniversal.bootloader({
    ...,
    template: '<app></app>'
})
```

Currently, Angular Universal automatically wraps the output in `<html>` etc even if it wasn't requested in the template. This feature is particuarly relevent when including Angular 2 server-side rendered content in an existing server-rendered page, e.g., on ASP.NET.

MarkPieszak:
Fixes #347

### DIFF
--- a/modules/platform-node/node-document.ts
+++ b/modules/platform-node/node-document.ts
@@ -26,17 +26,15 @@ export function parseDocument (documentHtml: string): any {
     throw new Error('parseDocument needs to be a string to be parsed correctly');
   }
 
-  const doc = parse5.parse(documentHtml, { treeAdapter : parse5.treeAdapters.htmlparser2 });
+  let doc;
 
-
-  /*
   // Build entire doc <!doctype><html> etc
   if (documentHtml.indexOf('<html>') > -1 && documentHtml.indexOf('</html>') > -1) {
-    const doc = parser.parse(documentHtml);
+    doc = parse5.parse(documentHtml, { treeAdapter : parse5.treeAdapters.htmlparser2 });
   }
   // ASP.NET case : parse only the fragment - don't build entire <html> doc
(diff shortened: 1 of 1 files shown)


MarkPieszak:
So is this one not helpful to fix the bigger issue (with Core handling the head title meta properties itself?)
@gdi2290

modules/platform-node/node-document.ts
```diff
@@ -26,17 +26,16 @@ export function parseDocument (documentHtml: string): any {
     throw new Error('parseDocument needs to be a string to be parsed correctly');
   }
 
-  const doc = parse5.parse(documentHtml, { treeAdapter : parse5.treeAdapters.htmlparser2 });
+  let doc;
 
-
-  /*
   // Build entire doc <!doctype><html> etc
   if (documentHtml.indexOf('<html>') > -1 && documentHtml.indexOf('</html>') > -1) {
```
gdi2290: this is still a problem with html tags that have attributes like `<html ⚡>` or `<html lang="en">`

MarkPieszak: Totally right. Should I clean it up to test for just <html, maybe even look for doctype as well? @gdi2290

gdi2290: you can use a regexp looking for any case of doctype

gdi2290:
might be supported in ng4 
https://github.com/angular/angular/issues/13822

PR #629, opened 2016-11-21
fix(node-renderer): fixes [hidden] issue

### LINKED ISSUE #612: [hidden]="true" not added to dom node when rendered on server
**Note: for support questions, please use one of these channels:** https://github.com/angular/universal/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports. Also, **Preboot has moved** to https://github.com/angular/preboot - please make preboot-related issues there.
- **I'm submitting a ...**
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support request here, see note at the top of this template.
- **What modules are related to this Issue?**
- [ ] express-engine
- [ ] grunt-prerender
- [ ] gulp-prerender
- [ ] hapi-engine
- [ ] universal-next
- [x] universal
- [ ] webpack-prerender
- **Do you want to request a _feature_ or report a _bug_?**

bug
- **What is the current behavior?**

Using [hidden]="true" in template does not add "hidden" property in server-side rendered dom
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** by creating a github repo.

Using the latest angular universal version, simply try to use [hidden]="true" on any element inside template and render on server. Then in the rendered dom there is no "hidden" property present.
- **What is the expected behavior?**

"hidden" property should be added to dom node.
- **What is the motivation / use case for changing the behavior?**
## 
- **Please tell us about your environment:**
- Angular version: 2.1.0
- Browser: [all]
- Language: [TypeScript 2.0.3 ]
- OS:  [Mac OS X]
- Platform: [NodeJs]
- **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

It seems that the "hidden" property is set but during rendering it is not applied to the actual dom node.

MarkPieszak:
Closes #612

You no longer need to use `[attr.hidden]`, `[hidden]` can be used like
normally suggested in Angular docs.

Tested with Boolean (true|false), and String|emptyString.

```
true | String.length > 0 will hide it.
false | emptyString, no hidden attribute added.

```

### DIFF
--- a/modules/platform-node/node-renderer.ts
+++ b/modules/platform-node/node-renderer.ts
@@ -60,7 +60,7 @@ export class NodeDomRootRenderer implements RootRenderer {
 
 }
 
-export const ATTRIBUTES = {
+export const ATTRIBUTES: { [name: string]: string[]; } = {
   textarea: [
     'autocapitalize',
     'autocomplete',
@@ -224,7 +224,7 @@ export const ATTRIBUTES = {
 };
 
 // TODO(gdi2290): provide better whitelist above to alias setting props as attrs
-export const IGNORE_ATTRIBUTES = {
+export const IGNORE_ATTRIBUTES: { [key: string]: boolean; } = {
   'innerHTML': true,
   'hidden' : true
 };
@@ -448,7 +448,7 @@ export class NodeDomRenderer extends DomRenderer {
     return el;
   }
 
-  _isObject(val) {
+  _isObject(val: any) {
     if (val === null) {
       return false;
     }
@@ -520,7 +520,7 @@ export class NodeDomRenderer extends DomRenderer {
     return super.invokeElementMethod(location, methodName, args);
   }
 
-  _setDisabledAttribute(renderElement, _propertyName, propertyValue) {
+  _setDisabledAttribute(renderElement: any, _propertyName: any, propertyValue: any) {
     if (isPresent(propertyValue)) {
       if (propertyValue === true || propertyValue.toString() !== 'false') {
         return super.setElementAttribute(renderElement, 'disabled', 'disabled');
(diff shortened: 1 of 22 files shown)


modules/platform-node/node-renderer.ts
```diff
@@ -559,6 +566,18 @@ export class NodeDomRenderer extends DomRenderer {
     }
     return super.setElementAttribute(renderElement, propertyName, propertyValue);
   }
+
+  _setHiddenAttribute(renderElement: any, propertyName: any, propertyValue: any) {
+    if (isPresent(propertyValue)) {
+      // Follow Angular default behavior, true || string (of anything) will cause it to be hidden
+      if (propertyValue === true || propertyValue.toString().length) {
```
KostyaEsmukov: I guess the `propertyValue.toString().length` would convert boolean `false` to a string and evaluate to true, which seems unexpected here.

    > let propertyValue = false
    undefined
    > !!(propertyValue === true || propertyValue.toString().length)
    true // I expect false here
    > !!(propertyValue.toString().length)
    true
    > propertyValue.toString()
    'false'

Please correct me if I'm wrong. I would simplify this condition as just `(propertyValue)` -- which will match both `true` and a non-empty string.

MarkPieszak: Is it running incorrectly when you pass in `false`? I was pretty sure I had scenarios for that thought it was working

KostyaEsmukov: Yes, I think it is. It renders `hidden="false"`, which actually makes the element hidden.
https://plnkr.co/edit/t65ZO82WziEygEBNgAVR?p=preview

MarkPieszak: In that case it's a String of "false", if you pass any string at all it'll hide it (that seems to be the default functionality in angular Core as well).

But you're right if you pass an actual `boolean false` it'll hit that OR operator, turn it into a string anyway and length === 5 ! Good catch 👍

MarkPieszak: Just patched it in #663

Issue #604, opened 2016-10-23
reflect-metadata should be a peer dependency


waterfoul:
- **I'm submitting a ...**
- [X] bug reporttemplate.
- **What modules are related to this Issue?**
- [X] angular2-universal-polyfills
- **Do you want to request a _feature_ or report a _bug_?**
  Bug
- **What is the current behavior?**
  angular2-universal-polyfills has reflect-metadata as a dependency instead of a peer dependency. This causes things to get detached within reflect-metadata if any other package adds a different version of reflect-metadata as a dependency. Note: This does not cause problems in webpack because of the way it loads modules
- **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** by creating a github repo.
- **What is the expected behavior?**
  reflect-metadata is a peer dependancy
- **What is the motivation / use case for changing the behavior?**
  Adding other potentially mis-configured packages or updating reflect-metadata
- **Please tell us about your environment:**
- Angular version: all
- Browser: all
- Language: all
- OS:  all
- Platform: all
- **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

PR #632, opened 2016-11-28
fix(polyfills): Remove reflect-metadata from the polyfills

waterfoul:
- [X] The commit message follows our guidelines: https://github.com/angular/universal/blob/master/CONTRIBUTING.md#commit-message-format
- [X] Tests for the changes have been added (for bug fixes / features)
- [X] Docs have been added / updated (for bug fixes / features)
   * The need to manually require this module is part of the angular core documentation

- [X] universal-polyfills

Bug fix

#604 

No

* **Other information**:

Change the polyfills to include reflect-metadata as a peer dependancy. If two modules import different versions of this module it will cause some very confusing errors. By including it as a peer npm will output a warning upon detecting a version mismatch and will only install the newer of the mismatched versions.

Fixes angular/universal#604

### DIFF
--- a/modules/universal-polyfills/package.json
+++ b/modules/universal-polyfills/package.json
@@ -7,8 +7,7 @@
   "dependencies": {
     "ie-shim": "<INHERIT>",
     "es6-promise": "<INHERIT>",
-    "es6-shim": "<INHERIT>",
-    "reflect-metadata": "<INHERIT>"
+    "es6-shim": "<INHERIT>"
   },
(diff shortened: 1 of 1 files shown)


gdi2290:
it's probably best just to drop support for the polyfill module. It was only meant to be convenient for demos and new repos but it also hides the polyfills. There is also a problem if you prefer core-js so it's best just to not provide the module.

gdi2290:
if you update the branch I can merge it

PR #635, opened 2016-12-08
docs(readme): update universalcache gotcha

MarkPieszak:
Updated to fix link:

 - Use a [UniversalCache](https://github.com/angular/universal-starter/blob/master/src/%2Bapp/shared/model/model.service.ts#L34-L50) instead of regular Http, to save certain requests so they aren't re-ran again on the Client. ([Example useage here](https://github.com/angular/universal-starter/blob/cc71e2d5b2d783f2bb52eebd1b5c6fa0ba23f08a/src/%2Bapp/%2Bhome/home.component.ts#L22-L24))

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ constructor(element: ElementRef, renderer: Renderer) {
 }
 ```
  - The application runs XHR requests on the server & once again on the Client-side (when the application bootstraps)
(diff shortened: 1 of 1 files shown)


MarkPieszak:
Any idea what the TravisCI is talking about here? @gdi2290

PR #681, opened 2017-03-21
feat(ng-express-engine): add ng-express-engine

FrozenPandaz:

- [x] ng-express-engine

feature

There is no express engine

There will be an express engine which allows server side rendering of Angular apps.

No

### DIFF
--- a/modules/ng-express-engine/README.md
+++ b/modules/ng-express-engine/README.md
@@ -1,0 +1,71 @@
+# Angular Express Engine
+
+This is an Express Engine for running Angular Apps on the server for server side rendering.
+
+## Usage
+
+To use it, set the engine and then route requests to it
+
+```ts
+import * as express from 'express';
+import { ngExpressEngine } from '@universal/ng-express-engine';
+
+// Set the engine
+app.engine('html', ngExpressEngine({
+  bootstrap: ServerAppModule // Give it a module to bootstrap
+}));
+
+app.set('view engine', 'html');
+
+app.get('/**/*', (req: Request, res: Response) => {
+  req: req,
+  res: res
+});
+```
+
+## Extra Providers
+
+Extra Providers can be provided either on engine setup
+
+```ts
+app.engine('html', ngExpressEngine({
+  bootstrap: ServerAppModule,
+  providers: [
+    ServerService
+  ]
+}));
+```
+
+## Advanced Usage
+
+### Request based Bootstrap
+
+The Bootstrap module as well as more providers can be passed on request
+
+```ts
+app.get('/**/*', (req: Request, res: Response) => {
+  req: req,
+  res: res,
+  bootstrap: OtherServerAppModule,
+  providers: [
+    OtherServerService
+  ]
+});
+```
+
+### Using the Request and Response
+
+The Request and Response objects are injected into the app via injection tokens.
+You can access them by @Inject
+
+```ts
+import { Request } from 'express';
+import { REQUEST } from '@universal/ng-express-engine';
+
+@Injectable()
+export class RequestService {
+  constructor(@Inject(REQUEST) private request: Request) {}
+}
+```
+
+If your app runs on the client side too, you will have to provide your own versions of these in the client app.
--- a/modules/ng-express-engine/index.ts
+++ b/modules/ng-express-engine/index.ts
@@ -1,0 +1,2 @@
+export { ngExpressEngine, NgSetupOptions, RenderOptions } from './src/main';
+export { RESPONSE, REQUEST } from './src/tokens';
--- a/modules/ng-express-engine/package.json
+++ b/modules/ng-express-engine/package.json
@@ -1,0 +1,47 @@
+{
+  "name": "@universal/ng-express-engine",
+  "main": "dist/main.js",
+  "types": "dist/index.d.ts",
+  "version": "1.0.0-beta.0",
+  "description": "Express Engine for running Server Angular Apps",
+  "homepage": "https://github.com/angular/universal",
+  "license": "MIT",
+  "contributors": [
+    "FrozenPandaz"
+  ],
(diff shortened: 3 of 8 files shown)


modules/ng-express-engine/index.ts
```diff
@@ -0,0 +1,101 @@
...
+            useValue: {
+              document: getDocument(filePath),
+              url: options.req.originalUrl
+            }
+          }
+        ]);
+
+      const moduleRefPromise = setupOptions.aot ?
+        platformServer(extraProviders).bootstrapModuleFactory(<NgModuleFactory<{}>>moduleFactory) :
+        platformDynamicServer(extraProviders).bootstrapModule(<Type<{}>>moduleFactory);
```
vikerman: Let's not provide a way to use platformDynamicServer. Instead let's compile the NgModule like we do here and cache the NgModuleFactory? - https://github.com/angular/angular/blob/master/packages/platform-server/test/integration_spec.ts#L377

modules/ng-express-engine/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "build": "tsc",
+    "prebuild": "rimraf dist"
+  },
+  "peerDependencies": {
+    "@angular/core": "^4.0.0-rc.5",
+    "@angular/platform-server": "^4.0.0-rc.5",
+    "express": "^4.15.2"
+  },
+  "devDependencies": {
+    "@angular/common": "^4.0.0-rc.5",
```
vikerman: I think we would also need @angular/animations since we now add a NoopanimationModule by default.

modules/ng-express-engine/index.ts
```diff
@@ -0,0 +1,101 @@
...
+ */
+function handleModuleRef(moduleRef: NgModuleRef<{}>, callback: Send) {
+  const state = moduleRef.injector.get(PlatformState);
+  const appRef = moduleRef.injector.get(ApplicationRef);
+
+  appRef.isStable
+    .filter((isStable: boolean) => isStable)
+    .first()
+    .subscribe(() => {
+      const bootstrap = moduleRef.instance['ngOnBootstrap'];
```
vikerman: ngOnBootstrap is not a standard way to get callbacks on bootstrap completion. APP_BOOTSTRAP_LISTENER is the way the app/module can get bootstrap completion callbacks.

modules/ng-express-engine/index.ts
```diff
@@ -0,0 +1,101 @@
...
+    } catch (e) {
+      callback(e);
+    }
+	}
+}
+
+function getReqResProviders(req: Request, res: Response): Provider[] {
+  const providers: Provider[] = [
+    {
+      provide: 'REQUEST',
```
vikerman: I am not sure whether we want to be in the business of directly injecting the request and response. The developer can send in extra providers in setupOptions and they can inject whatever is relevant to their domain - UserID etc. in whatever way they want.

modules/ng-express-engine/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+To use it, set the engine and then route requests to it
+
+```ts
+import * as express from 'express';
+import { ngExpressEngine } from '@universal/ng-express-engine';
+
+// Set the engine
+app.engine('html', ngExpressEngine({
+  bootstrap: ServerAppModule // Give it a module to bootstrap
```
vikerman: Please also specify how the user can send in extra providers

modules/ng-express-engine/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+app.get('/**/*', (req: Request, res: Response) => {
+  req: req,
+  res: res
+});
+```
+
+## Bootstrap
+
+The engine also calls the ngOnBootstrap lifecycle hook of the module being bootstrapped
```
vikerman: ngOnBootstrap does not seem to be an Angular concept. We shouldn't be adding anything new here. Apps can use the APP_BOOTSTRAP_LISTENER instead.

modules/ng-express-engine/index.ts
alxhub: We also _definitely_ don't want to be using strings to inject them, as that will be impossible after the deprecated forms of `Injector.get` go away. Typed tokens are much nicer.

modules/ng-express-engine/index.ts
alxhub: +1. It's not even on-bootstrap since it happens on stability which comes after bootstrap. Let's standardize this in 4.1, not make our own standard here.

modules/ng-express-engine/package.json
FrozenPandaz: This is for 4.0 release right?

modules/ng-express-engine/index.ts
FrozenPandaz: ngDoBootstrap is but its conditionally called as a backup for no bootstrap component.

I did the app bootstrap listener.. it doesnt look pretty in my opinion
https://github.com/FrozenPandaz/ng-universal-demo/commit/f2e21225e83788d1e2af5baf9d47bbe2fdaa814e

Should we have a ngOnStable?

modules/ng-express-engine/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+This is an Express Engine for running Angular Apps on the server for server side rendering.
+
+## Usage
+
+To use it, set the engine and then route requests to it
+
+```ts
+import * as express from 'express';
+import { ngExpressEngine } from '@universal/ng-express-engine';
```
alexweber: Is this right, a new `@universal` namespace?

MarkPieszak: Yup! `@universal/` will be used for all the engines and other add-ons to the new platform-sever. 🎈

PR #682, opened 2017-03-21
feat(ng-aspnetcore-engine): adding initial new engine

MarkPieszak:
WIP


modules/aspnetcore-engine/index.ts
```diff
@@ -0,0 +1,104 @@
...
+import 'rxjs/add/operator/first';
+
+export function ngAspnetCoreEngine(
+    providers: Provider[],
+    ngModule: Type<{}>
+): Promise<{ html: string, globals: { styles: string, title: string, meta: string, [key:string]: any } }> {
+
+    return new Promise((resolve, reject) => {
+
+        const platform = platformDynamicServer(providers);
```
vikerman: Please user platformServer only since platformDynamicServer would compile the whole app every time.

MarkPieszak: Yeah I need to do that and update the npm name as well.

PR #683, opened 2017-03-24
feat(ng-express-engine): support non-aot

FrozenPandaz:

- [x] Express Engine

remove unused proxy imports from some test files

Feature

Non-AoT Modules are not supported by the express engine

Non-AoT Modules will be compiled, cached and then used for the request.

Module Factories which have been cached will be used for subsequent requests.

Bootstrapped Module Factories will be used to requests like normal 

No

### DIFF
--- a/modules/ng-express-engine/src/file-loader.ts
+++ b/modules/ng-express-engine/src/file-loader.ts
@@ -1,0 +1,16 @@
+import * as fs from 'fs';
+import { ResourceLoader } from '@angular/compiler';
+
+export class FileLoader implements ResourceLoader {
+  get(url: string): Promise<string> {
+    return new Promise((resolve, reject) => {
+      fs.readFile(url, (err: NodeJS.ErrnoException, buffer: Buffer) => {
+        if (err) {
+          return reject(err);
+        }
+
+        resolve(buffer.toString());
+      });
+    });
+  }
+}
--- a/modules/ng-express-engine/src/main.ts
+++ b/modules/ng-express-engine/src/main.ts
@@ -1,16 +1,18 @@
 import * as fs from 'fs';
 import { Request, Response, Send } from 'express';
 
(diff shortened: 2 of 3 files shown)


modules/ng-express-engine/src/file-loader.ts
```diff
@@ -0,0 +1,10 @@
+import * as fs from 'fs';
+import { ResourceLoader } from '@angular/compiler';
+
+export class FileLoader implements ResourceLoader {
+  get(url: string): Promise<string> {
+    return new Promise((resolve) => {
+      resolve(fs.readFileSync(url).toString());
```
gdi2290: if you're using a promise you might as well use the async API

modules/ng-express-engine/src/main.ts
```diff
@@ -58,20 +74,52 @@ export function ngExpressEngine(setupOptions: NgSetupOptions) {
...
-      })
+      getFactory(moduleOrFactory, compiler)
+        .then(factory => {
+          return renderModuleFactory(factory, {
+            extraProviders: extraProviders
+          });
+        })
         .then((html: string) => {
           callback(null, html);
         });
```
gdi2290: you should also include an error callback to this promise

modules/ng-express-engine/src/main.ts
```diff
@@ -58,20 +74,57 @@ export function ngExpressEngine(setupOptions: NgSetupOptions) {
...
+        .then(factory => {
+          return renderModuleFactory(factory, {
+            extraProviders: extraProviders
+          });
+        })
         .then((html: string) => {
           callback(null, html);
+        }, (err) => {
+          callback(err);
+          throw err;
```
gdi2290: you shouldn't have to throw here since the callback is handling it

FrozenPandaz: I was doing this to pass along the error, but i guess they could do it in the callback

PR #690, opened 2017-04-06
fix(express-engine): Update package.json

vikerman:
- To new package namespace.
- Use 4.0.0 @angular packages


MarkPieszak:
LGTM!

PR #702, opened 2017-04-25
fix(build); update build & publish scripts

MarkPieszak:


MarkPieszak:
@vikerman @FrozenPandaz Also fixed the naming from `@nguniversal` to `@ng-universal`.
We should wipe out that older one before anyone uses it and point to the ng- dash one.

modules/ng-express-engine/README.md
```diff
@@ -4,13 +4,13 @@ This is an Express Engine for running Angular Apps on the server for server side
 
 ## Usage
 
-`npm install @nguniversal/express-engine --save`
+`npm install @ng-universal/express-engine --save`
```
vikerman: It is @nguniversal. Revert this one also.

MarkPieszak: Fixed

Issue #733, opened 2017-06-04
Accept compiler providers for the express engine


demurgos:
* **I'm submitting a ...**
```
- [ ] bug report
- [x] feature request
- [ ] support request => Please do not submit support request here, see note at the top of this template.
```

* **What modules are related to this Issue?**
```
- [ ] aspnetcore-engine
- [x] express-engine
- [ ] hapi-engine
```

* **Do you want to request a *feature* or report a *bug*?**

I want to request a *feature*.

* **What is the current behavior?**

It is currently impossible to pass providers for the compiler.

* **What is the expected behavior?**

It should be possible to pass providers used by the compiler.

* **What is the motivation / use case for changing the behavior?**

The i18n providers (`TRANSLATIONS`, `TRANSLATIONS_FORMAT` and `LOCALE_ID`) must be provided to the compiler. It is currently impossible to perform JIT internationalization following [the method of the cookbook](https://angular.io/docs/ts/latest/cookbook/i18n.html).

* **Please tell us about your environment:**

- Angular version: 4.1
- Browser: all
- Language: TypeScript 2.3
- OS:  Linux
- Platform: NodeJs

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

You could either pass the providers directly to the compiler factory or have a separated `compilerProviders` option.

Toxicable:
As mentioned in that PRI don't believe this is required for i18n right now

PR #746, opened 2017-06-23
feat(ng-express-engine): add support for compiler providers

demurgos:
```
```

```
- [ ] aspnetcore-engine
- [x] express-engine
- [ ] hapi-engine
```

This PR adds the `compilerProviders` setup option to to add custom providers to
the compiler used by the express engine, for example for i18n.

It is not possible use compiler providers, see #733.

It is now possible to use the `compilerProviders` setup option to use compiler providers.

No

* **Other information**:

I've been using this patch without any issue for a few projects but I'd like someone to check if it is OK.

### DIFF
--- a/modules/express-engine/README.md
+++ b/modules/express-engine/README.md
@@ -40,6 +40,9 @@ app.engine('html', ngExpressEngine({
 }));
 ```
 
+If you need to use providers for the compiler (for example for i18n), you may want to
+use the `compilerProviders` option.
+
 ## Advanced Usage
 
 ### Request based Bootstrap
--- a/modules/express-engine/src/main.ts
+++ b/modules/express-engine/src/main.ts
@@ -13,6 +13,7 @@ import { REQUEST, RESPONSE } from './tokens';
  */
 export interface NgSetupOptions {
   bootstrap: Type<{}> | NgModuleFactory<{}>;
+  compilerProviders?: StaticProvider[];
   providers?: StaticProvider[];
 }
 
@@ -40,11 +41,18 @@ const factoryCacheMap = new Map<Type<{}>, NgModuleFactory<{}>>();
 export function ngExpressEngine(setupOptions: NgSetupOptions) {
 
   const compilerFactory: CompilerFactory = platformDynamicServer().injector.get(CompilerFactory);
+
+  let compilerProviders: StaticProvider[] = [
+    { provide: ResourceLoader, useClass: FileLoader, deps: [] },
+  ];
+
+  if (setupOptions.compilerProviders) {
+    compilerProviders.push(...setupOptions.compilerProviders);
+  }
+
   const compiler: Compiler = compilerFactory.createCompiler([
     {
-      providers: [
-        { provide: ResourceLoader, useClass: FileLoader, deps: [] }
-      ]
+      providers: compilerProviders
     }
   ]);
 


demurgos:
I signed it!

FrozenPandaz:
This looks pretty good to me

modules/ng-express-engine/src/main.ts
```diff
@@ -40,11 +41,18 @@ const factoryCacheMap = new Map<Type<{}>, NgModuleFactory<{}>>();
 export function ngExpressEngine(setupOptions: NgSetupOptions) {
 
   const compilerFactory: CompilerFactory = platformDynamicServer().injector.get(CompilerFactory);
+
+  let compilerProviders: Provider[] = [
```
FrozenPandaz:
```ts
const compilerProviders: Provider[] = [
```

modules/ng-express-engine/src/main.ts
```diff
@@ -40,11 +41,18 @@ const factoryCacheMap = new Map<Type<{}>, NgModuleFactory<{}>>();
...
+    { provide: ResourceLoader, useClass: FileLoader },
+  ];
+
+  if (setupOptions.compilerProviders) {
+    compilerProviders = compilerProviders.concat(setupOptions.compilerProviders);
```
FrozenPandaz:
```ts
compilerProviders.push(setupOptions.compilerProviders);
```

demurgos:
Thank you for your review. I'll update my branch to use a `const` declaration.

As I explained in #733, my use case is to support i18n. While implementing it I noticed that compiler providers are effectively needed, but this first implementation can only be used to set the providers once due to cache (so one instance of Node cannot serve multiple languages). Since then I added finer control for cache to solve my problem. Should I add it to this PR or do a separate PR?

FrozenPandaz:
Good catch, please add it to this pr. 

_Sent from my Google Pixel XL using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

demurgos:
Hi,
I added a new commit to address your comments and add support for the cache control. My solution was to expose the cache key to the user so he can provide its own value. The engine was using the reference for the bootstrapped module previously (it's still the default value).
I also ensured that compiler providers are also used when defined in the options of a specific route handler (instead of only the setup options).

It turned out to be a bit larger PR so I tried to comment it. Let me know what you think about it.

demurgos:
@FrozenPandaz

Toxicable:
I don't think the cache control commit should be included in this PR.
See https://github.com/angular/universal/issues/811 and related issue for more info

demurgos:
Sorry for the long delay.
I rebased my commit adding only compiler providers (without cache) on top of `master`.

FrozenPandaz:
@ocombe would this be required to allow i18n for multiple languages?

ocombe:
I don't think so, you can already do that with universal since you only translate the templates of the generated views

FrozenPandaz:
@ocombe would that involve generating multiple bundles similar to what is done on the browser?

ocombe:
right now yes, in the future with the new renderer no because translations will be applied at runtime

demurgos:
These compiler providers (with my previous cache-control implementation) allows me to handle multiple languages with a single instance/bundle since May 2017.
If an official way of supporting this is added, I'm all for this but I believe that compiler providers may still provide some value in other context.

Toxicable:
@demurgos The only reason we're not just merging this is because we don't want to support public api if we don't have a solid use case for it

Toxicable:
I'm going to close this for now since it doesn't appear that core needs this to handle i18n

PR #776, opened 2017-08-10
fix(express-engine): adds deps to compilerFactory

CaerusKaru:
- Adds compatibility with latest version of Angular (v5.0.0-beta.3)
- Fixes #774

* **Please check if the PR fulfills these requirements**
```
- [x] The commit message follows our guidelines: https://github.com/angular/universal/blob/master/CONTRIBUTING.md#commit-message-format
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)
```

* **What modules are related to this pull-request**
```
- [ ] aspnetcore-engine
- [x] express-engine
- [ ] hapi-engine
```

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

Bug fix

* **What is the current behavior?** (You can also link to an open issue here)

#774 

* **What is the new behavior (if this is a feature change)?**

Adds compatibility with latest version of Angular (v5.0.0-beta.3)

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

No

PR #779, opened 2017-08-18
fix(aspnetcore-engine): add support for `StaticInjector`

Toxicable:
This is the same as #776 but for the aspnetcore-engine


vikerman:
Folded into https://github.com/angular/universal/commit/5c705cb51362bd448c0b8ac293da867a3a1913f4

PR #818, opened 2017-10-10
fix(build): require node 6 or above

vikerman:


vikerman:
Opened a bug for the CHANGELOG. We never did anything with it since Angular 4

PR #823, opened 2017-10-10
feat(common): add a common module with TransferHttpCacheModule

vikerman:
Add a TransferHttpCacheModule that when included in the App module will start caching Http requests on the server and transfer the cache to the client.

This will help avoid duplicate requests from the client for requests that the server application already made.

### DIFF
--- a/modules/common/src/transfer_http.ts
+++ b/modules/common/src/transfer_http.ts
@@ -1,0 +1,100 @@
+import {HTTP_INTERCEPTORS, HttpEvent, HttpHandler, HttpHeaders, HttpInterceptor, HttpRequest, HttpResponse} from '@angular/common/http';
+import {ApplicationRef, Injectable, NgModule} from '@angular/core';
+import {Observable} from 'rxjs/Observable';
+import {of} from 'rxjs/observable/of';
+import {filter} from 'rxjs/operator/filter';
+import {first} from 'rxjs/operator/first';
+import {toPromise} from 'rxjs/operator/toPromise';
+import {_do} from 'rxjs/operator/do';
+
+import {BrowserTransferStateModule, TransferState, makeStateKey} from '@angular/platform-browser';
+
+export interface TransferHttpResponse {
+  body?: any | null;
+  headers?: {[k: string]: string[]};
+  status?: number;
+  statusText?: string;
+  url?: string;
+}
+
+function getHeadersMap(headers: HttpHeaders) {
+  const headersMap: {[name: string]: string[]} = {};
+  for (const key of headers.keys()) {
+    headersMap[key] = headers.getAll(key)!;
+  }
+  return headersMap;
+}
+
+@Injectable()
+export class TransferHttpCacheInterceptor implements HttpInterceptor {
+
+  private isCacheActive = true;
+
+  private invalidateCacheEntry(url: string) {
+    this.transferState.remove(makeStateKey<TransferHttpResponse>('G.' + url));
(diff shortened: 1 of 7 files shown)


modules/common/src/transfer_http.ts
```diff
@@ -0,0 +1,105 @@
+import 'rxjs/add/observable/of';
```
Toxicable: any reason to not replace this with `of` import?

modules/common/src/transfer_http.ts
```diff
@@ -0,0 +1,105 @@
...
+  private invalidateCacheEntry(url: string) {
+    this.transferState.remove(makeStateKey<TransferHttpResponse>('G.' + url));
+    this.transferState.remove(makeStateKey<TransferHttpResponse>('H.' + url));
+  }
+
+  constructor(appRef: ApplicationRef, private transferState: TransferState,
+              @Inject(PLATFORM_ID) private platformId: Object) {
+    // Stop using the cache if the application has stabilized, indicating initial rendering is
+    // complete.
+    toPromise
```
Toxicable: Shouldn't this be wrapped in a `if(isPlatformBrowser())`?

patrickmichalina:
Glad this will become a common part of the npm package..

I'd recommend some tests.. We had a large amount of bugs in our app until we wrote all the tests for the http transfer..

https://github.com/patrickmichalina/fusebox-angular-universal-starter/blob/master/src/client/app/shared/transfer-http/transfer-http-interceptor.service.spec.ts

Toxicable:
@patrickmichalina We don't currently have a testing suite setup but once we do all modules will have tests added

modules/common/src/transfer_http.ts
vikerman: No. Both the server and client stop after initial render.

Toxicable: Then wouldn't it be better to use the onDoneRender ( I don't remember the actual name) token instead of onStable?

modules/common/src/transfer_http.ts
vikerman: No. Will fix it.

modules/common/src/transfer_http.ts
vikerman: The idea is that the cache is active only for the time of the initial render - Initial render is done when the app has become stable the first time.

Currently we don't want to be in the business of a generic HTTP cache past the initial render. But will decide it based on how people want to use it.

vikerman:
@patrickmichalina - Totally agreed on tests. I am in the process of setting up the tests in the CI - But wanted to get this out meanwhile.

modules/common/src/transfer_http.ts
vikerman: Hmm - This seems to be the canonical way to include Observable.of. What did you have in mind?

Toxicable: `import {of } from 'rxjs/observable/of';`
see here for example https://github.com/angular/angular/blob/3a227a1f6fb9d734aeba83cdf80246ae775a4f53/packages/router/src/utils/collection.ts#L12

modules/common/src/transfer_http.ts
Toxicable: ah right, that makes sense, also means slightly less code which is always nice

modules/common/src/transfer_http.ts
vikerman: Thanks - Done

naveedahmed1:
Is it available for aspnetcore-engine as well?

PR #825, opened 2017-10-10
feat(aspnet): update engine for angular 5.0 & domino adapter

MarkPieszak:
```
- [x] aspnetcore-engine
```

* **What kind of change does this PR introduce?** 
**feature** Update to 5.x & domino

No, I kept the engine in-tact so that current users only need to update all their dependencies, and everything will work as it did before.


modules/aspnetcore-engine/src/main.ts
```diff
@@ -1,19 +1,80 @@
...
+    UniversalData.links = LINKS.join(' ');
+  };
+}
 
-import 'rxjs/add/operator/filter';
-import 'rxjs/add/operator/first';
 
 export function ngAspnetCoreEngine(
   options: IEngineOptions
 ): Promise<{ html: string, globals: { styles: string, title: string, meta: string, transferData?: {}, [key: string]: any } }> {
```
Toxicable: transferData looks like it's unused, is this from before we had `TransferState`?

Toxicable:
Looks good, needs a commit squash though

PR #889, opened 2018-02-07
fix(aspnetcore-engine): fix ServerTransferStateModule and TransferState

stephenlautier:
```
```

```
- [x] aspnetcore-engine
- [ ] express-engine
- [ ] hapi-engine
```

Fix the ability to use `TransferState` and `ServerTransferStateModule`

When using `ServerTransferStateModule` it was breaking html from being rendered on the server. 

`ServerTransferStateModule` now work out of the box, and also returning the `moduleRef` which consumers can use to do some other logic with their services before transferring it to .net.

No breaking changes have been done, unless for some reason the consumer was using directly `UniversalData.title` (which is not even exposed)

* **Other information**:


Flood:
Great job Stephen!

MarkPieszak:
This looks great Stephen! @stephenlautier 
How come we need platform-server-utils, is this temporary until we get an actual fix into Core itself there?

stephenlautier:
@MarkPieszak I needed that to return the `moduleRef` to get the injector from it, ideally its just temp and it will be done in `angular/platform-server` i had opened an issue https://github.com/angular/angular/issues/22049

Unfortunately, I had never contributed to it, its quite complex, and I use windows some tooling doesn't seems to work well.

If anyone can help on that issue, it would be awesome!

PR #891, opened 2018-02-15
feat(build): add unit testing with karma

Toxicable:
Adds karma test runner


CaerusKaru:
Karma runs in the browser, is that what we want here? Aren’t we testing if these modules work on the server?

Toxicable:
Since they're very much related we should add both of them in the same PR to avoid conflicts, such as what spec files are being handled by what test runner. So I'll leave this up as reference for now while we organize the Jasmine one.

PR #896, opened 2018-02-16
chore(ci): clean up old config from travis

Toxicable:
cleaning up some old config from travis

I wasn't sure what these were, let me know if they're important
>env:
  global:
  - SAUCE_ACCESS_KEY=''
  - CXX=g++-4.8

cuts off about 2 min on build time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,9 @@ node_js:
   - '6'
   - '7'
   - '8'
+  - '9'
 
-cache:
-  directories:
-    - node_modules
-    - typings
+cache: yarn
 
 notifications:
   webhooks:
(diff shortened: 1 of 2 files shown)


CaerusKaru:
The SAUCE_ACCESS_KEY would be used if we planned on running unit tests with SauceLabs (which offers the ability to run them with different browsers and OS configurations). But, we probably don’t have much use for it, and Material is moving away from it because it’s so flaky, so it’s ok to remove IMO.

Toxicable:
@CaerusKaru Cool, well im happy removing that for now, we'll add it back in later if we need it. Or maybe look at alternatives if it's so flakey

CaerusKaru:
Sounds good. Material uses BrowserStack, but we should probably just get everything working on Travis first 😄

Toxicable:
Superseeded

PR #897, opened 2018-02-16
feat(modules): pass in absolute URL as default for view rendering

CaerusKaru:
* Add the absolute URL to view generation to allow for access through
  the document's native location object
* Future PR will allow user to override the default url

NOTE: This will act as a precursor to correct absolute path resolution for `HttpClient`

### DIFF
--- a/modules/aspnetcore-engine/src/main.ts
+++ b/modules/aspnetcore-engine/src/main.ts
@@ -128,7 +128,7 @@ export function ngAspnetCoreEngine(options: IEngineOptions): Promise<IEngineRend
         .then(factory => {
           return renderModuleFactory(factory, {
             document: options.document || options.appSelector,
-            url: options.url || options.request.url,
+            url: options.url || options.request.absoluteUrl,
             extraProviders: extraProviders
           });
         })
--- a/modules/express-engine/src/main.ts
+++ b/modules/express-engine/src/main.ts
@@ -76,6 +76,7 @@ export function ngExpressEngine(setupOptions: NgSetupOptions) {
 
       setupOptions.providers = setupOptions.providers || [];
 
+      const req = options.req;
       const extraProviders = setupOptions.providers.concat(
         options.providers,
         getReqResProviders(options.req, options.res),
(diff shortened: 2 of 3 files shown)


Toxicable:
I think we should hold off on this till we have a concrete solution for how HttpClient will injest the base url

CaerusKaru:
@Toxicable I thought about that, but I think we can consider this a separate issue. Is there ever a case we want the URL passed in to _not_ be absolute on the server?

Toxicable:
Blocked on deciding what we should do in core first

modules/express-engine/src/main.ts
```diff
@@ -84,7 +85,7 @@ export function ngExpressEngine(setupOptions: NgSetupOptions) {
             provide: INITIAL_CONFIG,
             useValue: {
               document: options.document || getDocument(filePath),
-              url: options.url || options.req.originalUrl
+              url: options.url || req.protocol + '://' + req.get('host') + req.originalUrl
```
Toxicable: also needs null check

modules/hapi-engine/src/main.ts
```diff
@@ -61,11 +62,12 @@ export function ngHapiEngine(options: RenderOptions) {
...
   ]);
 
-  if (options.req.raw.req.url === undefined) {
+  if (req.raw.req.url === undefined) {
     return Promise.reject(new Error('url is undefined'));
   }
 
-  const filePath = <string> options.req.raw.req.url;
+  const filePath = <string> req.raw.req.url;
+  const url = `${req.connection.info.protocol}://${req.info.host}${req.url.path}`;
```
Toxicable: last null check

modules/aspnetcore-engine/src/main.ts
```diff
@@ -128,7 +128,7 @@ export function ngAspnetCoreEngine(options: IEngineOptions): Promise<IEngineRend
         .then(factory => {
           return renderModuleFactory(factory, {
             document: options.document || options.appSelector,
-            url: options.url || options.request.url,
+            url: options.url || options.request.absoluteUrl,
```
Toxicable: needs null check on absoluteUrl

CaerusKaru: Not needed here

PR #898, opened 2018-02-17
build: switch to Angular Package Format

CaerusKaru:
* Restructure project to follow the Angular Package Format v5
* Update modules to conform to new structure
* Fix minor RxJS bugs
* Add Gulp to structure the workflow
* Bump deps for Angular/TypeScript to v5.2 and v2.6
* Add better Travis integration and new Travis tasks
* Add Jasmine scaffolding for server-side unit testing


CaerusKaru:
Note: this is a very big PR because it completely reorganizes the entire repo. I've split it into three commits:

1) add the Gulp utilities that perform the bundling. This is almost exactly pulled from Material and Flex Layout
2) update the modules so that they build correctly. This is mostly just tweaking the structure and adding necessary files
3) add Travis integration to automate this

Toxicable:
Instead of managing our own custom build system why not use something standard like ng-packagr? Especially since similar options are coming to the CLI soon, and we'll go away with this custom stuff then anyway

CaerusKaru:
1. The standard in the Angular organization is building packages using these proposed methods
2. We will not be managing this ourselves, the bulk of the maitenance work is done by the Material team, and we can just patch as necessary (these utilities were going to be spun off into their [own repo](https://github.com/angular/material2-build-tools))
3. Until a timeline has been announced for the CLI integration, and similar plans are available for the progression of the other Angular repos, it is a mistake to wait while these utilities are available, and working, now
4. When those utilities become available, it will be far easier to simply duplicate the transition that Material makes in their bundling process than to come up with a new one for us from scratch

CaerusKaru:
Closed in favor of #903

PR #903, opened 2018-02-22
test: add testing setup for Jasmine, clean up Karma

CaerusKaru:
* Add testing mode for server
* Clean up testing mode for browser
* Bump deps to Angular v5.2 and TypeScript v2.6
* In order to land the testing changes, it was necessary to upgrade
  the build setup. This is in anticipation of the switch to Bazel
  and incorporates many of the changes that will need to be made
  then


.travis.yml
```diff
@@ -1,43 +1,23 @@
...
-    on_start: false     # default: false
-  slack:
-    secure: EP4MzZ8JMyNQJ4S3cd5LEPWSMjC7ZRdzt3veelDiOeorJ6GwZfCDHncR+4BahDzQAuqyE/yNpZqaLbwRWloDi15qIUsm09vgl/1IyNky1Sqc6lEknhzIXpWSalo4/T9ZP8w870EoDvM/UO+LCV99R3wS8Nm9o99eLoWVb2HIUu0=
-
-env:
-  global:
-  - SAUCE_ACCESS_KEY=''
-  - CXX=g++-4.8
+node_js:
+  - '8.9.4'
```
Toxicable: update package.json to reflect this

.travis.yml
```diff
@@ -1,43 +1,23 @@
...
+  - master
 
-before_script:
+install:
   - npm install
 
 script:
   - ./build.sh
 
+cache:
```
Toxicable: should be `cache: yarn`

.travis.yml
```diff
@@ -1,43 +1,23 @@
...
-before_install:
-  # Updating NPM to relevant version >= 3 on Node.JS LTS
-  - npm i -g npm@^3
+branches:
+  only:
+  - master
 
-before_script:
+install:
   - npm install
```
Toxicable: should be `yarn install --frozen-lockfile`

build-test.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const path = require('path');
+const ngc = require('@angular/compiler-cli/src/main').main;
+const rimraf = require('rimraf');
+
+const rootFolder = path.join(__dirname);
+const distFolder = path.join(rootFolder, 'dist');
+
+const libNames = [
```
Toxicable: THese names should be centralized somewhere else

build.js
```diff
@@ -0,0 +1,206 @@
...
+  'rxjs/operators/filter': 'Rx.operators',
+  'rxjs/operators/map': 'Rx.operators',
+  'rxjs/operators/take': 'Rx.operators',
+  'rxjs/operators/tap': 'Rx.operators',
+  'rxjs/observable/of': 'Rx.Observable',
+  'fs': 'fs'
+};
+
+const namespace = '@nguniversal';
+const libNames = [
```
Toxicable: as prior, shouldn't be duplicated

build.js
```diff
@@ -0,0 +1,206 @@
+'use strict';
```
Toxicable: I don't think we should be doing this file.
Either use ng_packger since that already does APF or leave it as is till we put ng-package in

build.sh
```diff
@@ -1,28 +1,3 @@
...
-npm run build:hapi-engine || exit 1
-
-cp modules/hapi-engine/package.json dist/hapi-engine/package.json
-cp modules/hapi-engine/README.md dist/hapi-engine/README.md
-
-npm run build:module-map-ngfactory-loader || exit 1
-
-cp modules/module-map-ngfactory-loader/package.json dist/module-map-ngfactory-loader/package.json
-cp modules/module-map-ngfactory-loader/README.md dist/module-map-ngfactory-loader/README.md
+npm run build
```
Toxicable: this is pointless

CaerusKaru: It's needed so that we can run it from Travis, at least until we get a full Travis script in a follow-up PR.

build.js
CaerusKaru: As mentioned, `ng_packagr` is not ready yet, and at the very least this is a fallback in case it takes longer/has bugs.

It's very low impact and does exactly what we need to right now.

Toxicable: Sorry I meant those the other way around.
This is what I mean: https://github.com/dherges/ng-packagr

test/test-browser.ts
```diff
@@ -0,0 +1,15 @@
+// Load karma not outside. Karma pollutes Promise with a different implementation.
```
Toxicable: Any reason why we need a file to do this? why not jsut run from cli?

build.js
CaerusKaru: How is adding a whole dependency better than this?

test/remote-browsers.json
```diff
@@ -0,0 +1,277 @@
+{
```
Toxicable: What's the purpose of these browsers?

CaerusKaru: When we eventually integrate browser testing in Travis, this is a list of preset browsers we can draw on.

build.js
Toxicable: We don't have to manage it ourselves and it already implements APF for us

test/test-browser.ts
CaerusKaru: To maintain parity with the browser version and allow for additional configuration

test/remote-browsers.json
Toxicable: We should do that when we do that integration then, no need to add config we're not using right now.

build.js
CaerusKaru: This is pulled from the CLI's initial library packaging system. It's in use already in preboot. It's one file. It's a temporary solution until we switch to Bazel. It's not a whole additional dependency that we then have to manage and keep tabs on.

test/remote-browsers.json
CaerusKaru: No reason we can't, and it's in use in the Karma config

test/test-browser.ts
Toxicable: Config can be done from the config file, also you're passing that config file in so you cant edit it.
Also that dosen't appear to be able to run watch mode

CaerusKaru: That's the whole point. Running from the command line should be auto-watch. From the runner, it should be single-run.

build.js
Toxicable: The CLI will be using ng-packgr to build libs in v6. So I don't think that's a reasonable concern.

build.sh
Toxicable: huh? you can have travis run a specific npm script

modules/common/tsconfig.lib.json
```diff
@@ -0,0 +1,37 @@
...
+    "inlineSources": true,
+    "target": "es2015",
+    "lib": ["es2015", "dom"],
+    "skipLibCheck": true,
+    "types": [],
+    "baseUrl": "."
+  },
+  "files": [
+    "./public-api.ts",
+    "./typings.d.ts"
```
Toxicable: Why do we have typings files? We shouldn't be using any untyped libs

build.sh
CaerusKaru: News to me, fixed.

test/test-browser.ts
Toxicable: Oh so this is for CI only?

CaerusKaru: Yep!

.travis.yml
```diff
@@ -1,43 +1,21 @@
...
+  - master
 
-before_script:
-  - npm install
+install:
+  - yarn install --frozen-lockfile
 
 script:
-  - ./build.sh
+  - npm run build
```
Toxicable: add the testing scripts here, they aren't being run on Travis
Also add a example Jasmine script that does something similar to `expect(true).toBe(true)`

CaerusKaru: The Jasmine script runs the same thing as the Karma script. I've moved it into common for now as `transfer_http.spec.ts`

Toxicable:
Overall looks good for now.
Will take a better look this afternoon

modules/common/tsconfig.lib.json
CaerusKaru: Past precedent in other libs. I'll remove them for now and we can add them back if we need them. Sometimes it can't be helped btw.

Toxicable:
Your Karma script has some issue.
It cant start some browser then throws. This isn't casuing a failure

scripts/ci/env.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [[ -z "${TRAVIS}" ]]; then
+  echo "This script can only setup the environment inside of Travis builds"
+  exit 0
+fi
+
+# - we overwrite the value set by Travis JWT addon here to work around travis-ci/travis-ci#7223 for FIREBASE_ACCESS_TOKEN
+export SAUCE_ACCESS_KEY=9b988f434ff8-fbca-8aa4-4ae3-35442987
```
Toxicable: uneeded?

scripts/ci/travis-env.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
Toxicable: same as this? what do we need this for?

scripts/ci/sources/tunnel.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Load the retry-call utility function.
+source ./scripts/retry-call.sh
+
+# Variable the specifies how often the wait script should be invoked if it fails.
+WAIT_RETRIES=2
+
+start_tunnel() {
+  case "$MODE" in
+    e2e*|saucelabs*)
```
Toxicable: Is this needed?

scripts/ci/travis-script.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
Toxicable: I don't think this is needed either

scripts/retry-call.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
Toxicable: Also no needed

scripts/ci/travis-script.sh
CaerusKaru: I kept this file but deleted the "travis-testing" script (and replaced the contents of this one)

package.json
```diff
@@ -54,45 +54,50 @@
...
-    "build:common": "ngc -p modules/common/tsconfig.json",
-    "build:express-engine": "ngc -p modules/express-engine/tsconfig.json",
-    "build:aspnetcore-engine": "ngc -p modules/aspnetcore-engine/tsconfig.json",
-    "build:hapi-engine": "ngc -p modules/hapi-engine/tsconfig.json",
-    "build:module-map-ngfactory-loader": "ngc -p modules/module-map-ngfactory-loader/tsconfig.json",
-    "build": "./build.sh",
-    "test:watch": "karma start karma.conf.js",
-    "test": "karma start karma.conf.js --single-run"
+    "build": "node build.js",
+    "test:watch": "node build-test.js && karma start karma.conf.js",
```
Toxicable: None of these look like they run in watch mode. That'll be pretty lame for dev

test/karma.conf.js
```diff
@@ -0,0 +1,66 @@
+const path = require('path');
```
Toxicable: Any particular reason for removing puppeteer?

CaerusKaru: We don't need it

package.json
CaerusKaru: `test:watch` runs in watch mode because auto-watch is on by default

Toxicable: Sure it runs karma in autowatch but it only builds the project once

test/karma.conf.js
Toxicable: then remove it from the package.json

package.json
CaerusKaru: Let's add that in a follow-up PR. This is really only supposed to be to get the skeleton in place

test/karma.conf.js
CaerusKaru: Removed

package.json
Toxicable: Aight sounds good

PR #912, opened 2018-02-25
feat(common): add UniversalEngine to encapsulate rendering

CaerusKaru:
* Abstract the common rendering patterns and caching into
  a UniversalEngine
* Update build parameters to reflect dependency on
  common for existing engines

BREAKING CHANGE:

* The ASP.NET Core, Express, and Hapi engines now have
  the Common package as a peer dependency
* The tokens for all engines are now exported from
  `@nguniversal/common`

E.g.

BEFORE:
`import {REQUEST, RESPONSE} from '@nguniversal/express-engine/tokens';`

AFTER:
`import {REQUEST, RESPONSE} from '@nguniversal/common';`


CaerusKaru:
Full credit to @Toxicable for getting this together. Removed from the commit history because the clabot is evil.

modules/aspnetcore-engine/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+// Configuration for IDEs only.
```
Toxicable: this was left over from the factor, should be removed

modules/aspnetcore-engine/tsconfig.lib.json
```diff
@@ -21,7 +21,10 @@
     "lib": ["es2015", "dom"],
     "skipLibCheck": true,
     "types": ["node"],
-    "baseUrl": "."
+    "baseUrl": ".",
+    "paths": {
+      "@nguniversal/common": ["../../dist/packages/common"]
```
Toxicable: Should point to `../common` - the src not the output, unless this was the fix to the issue I had

modules/common/src/universal-engine/file-loader.ts
```diff
@@ -8,6 +8,9 @@
 import * as fs from 'fs';
 import { ResourceLoader } from '@angular/compiler';
 
+/**
+ * ResourceLoader implimentation for loading files
+ */
 export class FileLoader implements ResourceLoader {
```
Toxicable: Should be `@Interval ` along with the other ones under UniversalEngine

modules/common/src/universal-engine/universal-engine.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+import {ResourceLoader} from '@angular/compiler';
+import {Compiler, Type, NgModuleFactory, CompilerFactory, StaticProvider} from '@angular/core';
+import {INITIAL_CONFIG, renderModuleFactory, platformDynamicServer} from '@angular/platform-server';
+import * as fs from 'fs';
+
+import {FileLoader} from './file-loader';
+import {RenderOptions} from './interfaces';
+
+export class UniversalEngine {
```
Toxicable: should be `@internal`

modules/common/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+// Configuration for IDEs only.
```
Toxicable: also unneeded

modules/express-engine/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+// Configuration for IDEs only.
```
Toxicable: same here

modules/aspnetcore-engine/tsconfig.json
CaerusKaru: I think it's important to note that this file should not be extended, and is not being used to build the package

modules/aspnetcore-engine/tsconfig.lib.json
CaerusKaru: This was exactly the fix to the issue you had

modules/common/src/universal-engine/file-loader.ts
CaerusKaru: This file is not exported in the public api so is by default internal

CaerusKaru: Unless you mean I should just add an internal comment? In which case sure

modules/common/src/universal-engine/universal-engine.ts
CaerusKaru: Same as above

modules/common/tsconfig.json
CaerusKaru: Same as above

modules/express-engine/tsconfig.json
CaerusKaru: Same as above

modules/aspnetcore-engine/tsconfig.json
Toxicable: What's it being used for then?

CaerusKaru: IDEs, as noted by the comment. It's so that the imports are not considered invalid while editing files.

modules/common/src/universal-engine/universal-engine.ts
CaerusKaru: Also, I don't think this should be internal. I think the whole point is to encourage usage, even outside of the nguniversal-scope engines.

Toxicable: That wasn't my original plan. I think for now we should keep it internal, we can always externalize it later on, we cannot internalize it once it's already external

CaerusKaru: Alright I've marked it and its methods internal for now, but we really should look at this soon to externalize

build-config.js
```diff
@@ -29,8 +29,8 @@ const namespace = 'nguniversal';
 
 /** The library entrypoints that are built under the namespace */
 const libNames = [
-  'aspnetcore-engine',
   'common',
+  'aspnetcore-engine',
```
Toxicable: any reason for this?

CaerusKaru: Not anymore, was leftover from when we used the pre-Material build system. The rationale being that common needed to be built before everything else

WORKSPACE
```diff
@@ -10,18 +10,20 @@ http_archive(
...
+    url = "https://github.com/bazelbuild/rules_nodejs/archive/0.6.0.zip",
+    strip_prefix = "rules_nodejs-0.6.0",
+    sha256 = "e8a2bb5ca51fbafb244bc507bcebcae33a63d969f47413b319a8dcce032845bf",
 )
 
 load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "yarn_install")
 node_repositories(package_json = ["//:package.json"])
 
 
-local_repository(
```
Toxicable: This is easier to leave as is since it'll use the one that Angular core depends on, otherwise we could get out of sync

CaerusKaru: Point taken

modules/common/public_api.ts
```diff
@@ -7,3 +7,6 @@
  */
 export { TransferHttpCacheModule } from './src/transfer_http';
 export { StateTransferInitializerModule } from './src/state-transfer-initilizer/module';
+export { UniversalEngine as ɵUniversalEngine } from './src/universal-engine/universal-engine';
+export { NgSetupOptions } from './src/universal-engine/interfaces';
+export { REQUEST, RESPONSE, ORIGIN_URL } from './src/tokens/index';
```
Toxicable: Tokens should be a secondary entry point here.
Otherwise we'll get size regressions

modules/common/src/universal-engine/index.ts
```diff
@@ -5,4 +5,5 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-export * from './public_api';
+export * from './interfaces';
+export {UniversalEngine as ɵUniversalEngine} from './universal-engine';
```
Toxicable: could you make a private export file for common and do the renaming there

modules/express-engine/BUILD.bazel
```diff
@@ -10,22 +10,18 @@ ng_module(
         "*.ts",
         "src/**/*.ts",
     ]),
-    module_name = "@nguniversal/express-engine",
```
Toxicable: put this back in

Toxicable:
There's a regression here with `TransferHttpCacheModule`
Since it's bundled into the same bundle as the Universal Engine, it causes the `TransferHttpCacheModule` import to bring domino and other server side symbols with it.

I propose we make a secondary entry for the Engine to avoid any more API changes

Toxicable:
We should reexport the tokens from their old locations but depricate them so this isn't a breaking change, to be removed in v7

CaerusKaru:
Closed in favor of #996

PR #922, opened 2018-03-10
build: add secondary entry points for tokens

Toxicable:
upgrades angular to v6-rc.1
uses bazel to do secondary entrypoints for tokens


Toxicable:
Ran into some issue with not being able to use out reexported `ng_module`
Working around it by using the one exported from angular instead.
But otherwise I think this is good for review, havn't tested if this actually solve the issue yet though, havn't had the time

Toxicable:
Looks like CI is getting some very unhelpful error.

modules/express-engine/src/main.ts
```diff
@@ -17,7 +17,8 @@ import {
 } from '@angular/platform-server';
 
 import { FileLoader } from './file-loader';
-import { REQUEST, RESPONSE } from './tokens';
+//  import { REQUEST, RESPONSE } from './tokens/index';
```
Toxicable: remove

modules/aspnetcore-engine/tokens/tokens.ts
```diff
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+
```
Toxicable: remove

modules/express-engine/tokens/tokens.ts
```diff
@@ -5,6 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
+
```
Toxicable: remove

modules/hapi-engine/BUILD.bazel
```diff
@@ -1,20 +1,37 @@
-load("//tools:defaults.bzl", "ts_library", "ng_module", "ng_package")
+load(
+    "//tools:defaults.bzl",
+    "ts_library",
+    # "ng_module",
```
Toxicable: add comment explaining this

modules/hapi-engine/tokens/tokens.ts
```diff
@@ -5,7 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import { Request, Response } from 'hapi';
+
```
Toxicable: remove this

axnsan12:
Hello,

I see that this is marked as a breaking change and milestoned for 6.0.  
I think this is wrong - at least for ``express-engine``, because I had sucesfully been using ``import {REQUEST} from '@nguniversal/express-engine/tokens';`` in ``"@nguniversal/express-engine": "^5.0.0-beta.5"``.  

However after "upgrading" to ``"@nguniversal/express-engine": "^5.0.0-beta.6",``, I was forced to change to ``import {REQUEST} from '@nguniversal/express-engine';`` which increased my bundle size by the mentioned 600kb.  

I don't think you should force us to wait for ``6.0`` for a fix to a regression introduced in a minor version.

Just my 2 cents, and thank you for your good work on this project!

Toxicable:
@axnsan12 sorry about the confusion, this is definitely going into 5.x, notice the PR Target 5.x & master.
Working on it as much as i can, should hopefully be released by end of week.

Toxicable:
Just a small update here.
I've made this fix with another PR here: https://github.com/angular/universal/pull/930
which has been merged, just awaiting a release.

Toxicable:
This is almost ready to go. 2 small issues:
im getting this
```

File: BUILD.bazel [hunk 6]
```diff
@@ -4,22 +4,41 @@ package(default_visibility = ["//visibility:public"])
 
 exports_files(["tsconfig.json"])
 
-# Provides the npm-installed dependencies to Bazel actions.
 filegroup(
```
alexeagle: there is now an (undocumented) `node_modules_filegroup` that you should probably use instead, just for readability: https://github.com/bazelbuild/rules_nodejs/blob/master/defs.bzl#L41

modules/hapi-engine/tokens/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "name": "@nguniversal/hapi-engine/tokens"
```
alexeagle: hmm, you shouldn't need to add extra `package.json` files - `ng_package` should generate these for secondary entry points as needed

tools/defaults.bzl
```diff
@@ -31,9 +39,30 @@ def ng_package(globals = None, **kwargs):
...
       "fs": "fs",
       "express": "express",
       "hapi": "hapi"
     }
-  _ng_package(globals = globals, **kwargs)
+
+def ts_test_library(node_modules=None, **kwargs):
+    ts_library(testonly=1, **kwargs)
+
+def ts_library(tsconfig = None, node_modules = None, **kwargs):
```
vikerman: Why are ts_library, ng_module and ng_package redefined here? I would rather prefer us specifying this everywhere by hand,

Inside Google we have learnt never to use macros except for final leaf targets and when completely unavoidable.

tools/defaults.bzl
```diff
@@ -31,9 +39,30 @@ def ng_package(globals = None, **kwargs):
       "@nguniversal/express-engine/tokens": "nguniversal.expressEngine.tokens",
       "@nguniversal/hapi-engine/tokens": "nguniversal.hapiEngine.tokens",
       "rxjs": "Rx",
-      "rxjs/operators": "Rx.operators"
+      "rxjs/operators": "Rx.operators",
```
vikerman: I don't completely understand what the GLOBALS is doing. But I'm scared seeing the symbol Rx here. Rx.ts in rxjs pulls in the entire library. Are we sure that's not happening here?

tools/defaults.bzl
Toxicable: While I agree it's not the best idea, it's so we don't have to redefine those variables and can keep them organised in one place since there is no alternative right now.
This is the same as Angular core https://github.com/angular/angular/blob/master/tools/defaults.bzl
Once we have a better way of specifying defaults then we wont need this.

tools/defaults.bzl
Toxicable: This is ensuring that the entire lib is not being pulled into the umd bundle. The symbol on the right is what it will be named on the global object when passed in.

tools/defaults.bzl
vikerman: Alright - Spoke with Alex and this is a temporary work around needed. Let's add a TODO comment that this should be revisited at some point.

Issue #920, opened 2018-03-09
How it import express tokens without increasing the bundle size


dottodot:
in express-engine@5.0.0-beta.5 the token where imported as 

```
import { REQUEST, RESPONSE } from '@nguniversal/express-engine/tokens';
```
but in express-engine@5.0.0-beta.6 you have to do 
```
import { REQUEST, RESPONSE } from '@nguniversal/express-engine';
```
This change alone adds domino into the production build adding an extra 280kb to the bundle size.

Toxicable:
Thanks  for reporting us. This was a side affect of our recent change to use the Angular Package Format.
We're currently working on a fix to use a secondary entrypoint (similar to what it was before).

Until then we suggest either going back to beta.5 for now.

Will update when we have a fix

PR #930, opened 2018-03-14
build: re-introduce build.sh

Toxicable:
BREAKING CHANGE: Tokens are now exported from @nguniversal/engineName/tokens
This change fixes are regression where using these tokens in your browser app would increase the bundle size by about ~600kb.

fixes: #920


modules/common/public-api.ts
```diff
@@ -6,4 +6,4 @@
  * found in the LICENSE file at https://angular.io/license
  */
 export { TransferHttpCacheModule } from './src/transfer_http';
-export { StateTransferInitializerModule } from './src/state-transfer-initilizer/module';
+// export { StateTransferInitializerModule } from './src/state-transfer-initilizer/module';
```
Toxicable: don't export this into v5 since it actually fails to compile and hasn't been validated.

modules/express-engine/tsconfig.fortokens.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./tsconfig.lib.json",
+  "files": [
+    "./index.ts",
```
CaerusKaru: Don't need the `./` here, or elsewhere

modules/express-engine/tsconfig.fortokens.json
```diff
@@ -0,0 +1,8 @@
...
+    "./tokens.ts"
+  ],
+  "angularCompilerOptions": {}
```
CaerusKaru: Remove

Toxicable: the `angularCompilerOptions` ? we need to overide them since the ones declared in `.lib` conflict with our `files` declaration here

CaerusKaru: They don't get extended, see https://github.com/angular/angular/issues/22684

Toxicable: ahh sweet as then, wasn't aware of that

modules/express-engine/tsconfig.fortokens.json
Toxicable: it was like that in the other tsconfigs so being consistent

CaerusKaru: Ok then that's fine, didn't realize it was like that

PR #942, opened 2018-03-23
fix(packaging): add better script for publishing

Toxicable:
So now to deploy you run
`yarn upvery:whatever kind of bump it is`
`yarn deploy`


Toxicable:
closing as this will be added to the publish script

PR #953, opened 2018-04-05
test(express-engine): add basic unit tests

Toxicable:
Some super basic unit testing for express-engine. But it shows a few issues so far.

### DIFF
--- a/modules/express-engine/BUILD.bazel
+++ b/modules/express-engine/BUILD.bazel
@@ -36,6 +36,7 @@ ts_library(
     ]),
     deps = [
         ":express-engine",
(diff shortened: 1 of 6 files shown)


CaerusKaru:
Scope is `test`, e.g. `test(express-engine): add unit tests`. I've marked this as WIP since it looks like a lot still needs to be worked out (namely the `xit` stuff and the marked TODOs). Please ping when this is ready for review.

Toxicable:
CI failures aren't actually breaking the build. It's becasue I linked it up with Bazel not the old stuff.

PR #977, opened 2018-05-04
test(e2e): add e2e testing

CaerusKaru:


CaerusKaru:
Blocked while lazy loading e2e test is being fixed

integration/run_tests.sh
```diff
@@ -0,0 +1,57 @@
...
+
+# basedir is the workspace root
+readonly basedir=$(pwd)/..
+
+# Track payload size functions
+if $CI; then
+  # We don't install this by default because it contains some broken Bazel setup
+  # and also it's a very big dependency that we never use except when publishing
+  # payload sizes on CI.
+  yarn add --silent -D firebase-tools@3.12.0
```
Toxicable: We don't need this

WORKSPACE
```diff
@@ -3,32 +3,34 @@ workspace(name = "nguniversal")
...
 
 load("@build_bazel_rules_nodejs//:defs.bzl", "check_bazel_version", "node_repositories", "yarn_install")
 
-check_bazel_version("0.11.1")
+check_bazel_version("0.13.0")
 node_repositories(package_json = ["//:package.json"])
 
 
-local_repository(
+http_archive(
```
Toxicable: it's easier to use the one that @angular/bazel pulls in rather than changing this on every version chance since it's easy toget out of sync with angular

CaerusKaru: So while that's a fine point in principle, here's what happens to me every time, without fail:

I end up needing to reinstall node_modules, but running `bazel run @yarn//:yarn` without node_modules will fail, unless this is an `http_archive`, because Bazel will look for this and not find it and error out.

I think since we're already tracking the other dependencies (nodejs, bazel itself), it's not that much more to also track this.

integration/run_tests.sh
CaerusKaru: Agreed, but we will once we start tracking payload sizes. We can't have an empty if, and this amounts to a no-op at the moment. The alternative is to just comment out the if/else for the time being, but I figured this would preserve the intention. Open to either or.

Toxicable: we wont need it for tracking file sizes

WORKSPACE
Toxicable: ah, I wasn't aware of those issues, i've never had them.
For that reason alone I think it's fair to track this dep

Toxicable:
@googlebot wow I see how it is!

Issue #927, opened 2018-03-11
TransferHttpCacheModule not sending correct request


dottodot:
There seems to be an issue with `TransferHttpCacheModule` sending the previous request to the server. When I navigating to and from different product categories I will some get the wrong response returned i.e I'll get results for the last category I visited rather than the one it supposed to return.

I can see the value from `route.paramMap` is changing but the server is not receiving that value it get the last value.

```
export class ShopListResolver implements Resolve<any> {
  constructor(private service: ApiService, private router: Router) {
    service.endpoint = 'productdisplay';
  }

  resolve(
    route: ActivatedRouteSnapshot,
    state: RouterStateSnapshot
  ): Observable<any> {
    const category = route.paramMap.get('category');
    return this.service
      .find({
        $category: category,
        published: true,
        $sort: ['weight'],
        $select: ['title', 'listing_image', 'product_type', 'slug']
      })
      .pipe(take(1));
  }
}
```

dottodot:
Is there any reason why `req.urlWithParams` couldn't be used rather than `req.url`

njl07:
You can use module from `Hapiness` framework which is an enhancement of initial module.

Documentation is [here](https://github.com/hapinessjs/ng-universal-transfer-http)

This module caches all **type of request** with **all parameters** and supports `sequential` and `parallel` calls.

I use it in several projects in production and all work very well.

After that combine it with [Hapiness module](https://github.com/hapinessjs/ng-universal-module) for `SSR`.

You can find an example of the implmentation [here](https://github.com/hapinessjs/ng-universal-example)

hiepxanh:
I do not understand why toxiccable is reject that PR. @CaerusKaru  I would like you to reconsider that PR and merge it. because I have problem with transferHttpCache module. I would like to cache on params and I have pure `graphql` application which make all multation and query on `POST` request. that PR is good idea. maybe we can have an option to have like:
```
{
   cacheMethod: ['GET'] // this is default options can be: ['GET', 'POST'] etc.....
   cacheParam: true // false is default
}
```
what do you think? I can make a PR if you allow me to modify? @CaerusKaru

hiepxanh:
@alan-agius4 can you make some comment?

alan-agius4:
The cache is based on params see: https://github.com/angular/universal/pull/1005 and https://github.com/angular/universal/blob/7429b4054716426e9c9c0ae2b50c68021ba87ea9/modules/common/src/transfer_http.ts#L59-L69

[...]

Issue #855, opened 2017-11-29
after 1 post request no get request is beeing cached


lutzk:
**Note: for support questions, please use one of these channels:** https://github.com/angular/universal/blob/master/CONTRIBUTING.md#question. This repository's issues are reserved for feature requests and bug reports. Also, **Preboot has moved** to https://github.com/angular/preboot - please make preboot-related issues there.

* **I'm submitting a ...**
```
- [x] bug report
- [ ] feature request
- [ ] support request => Please do not submit support request here, see note at the top of this template.
```

* **What modules are related to this Issue?**
```
- [ ] aspnetcore-engine
- [ ] express-engine
- [ ] hapi-engine
- [x] common
```

* **Do you want to request a *feature* or report a *bug*?**
bug i think - not sure

* **What is the current behavior?**

when i use the `TransferHttpCacheModule` and the `BrowserTransferStateModule` in an universal app and do a HttpClient.get the request and all later get requests to diffrent urls/ enpoints are cached in the TransferState and transfered to the browser from gthe server - so far so good

but if i do a HttpClient.post request none of the later requests is cached regardless if its a HttpClient.get or not

when i add a else block in
 [https://github.com/angular/universal/blob/master/modules/common/src/transfer_http.ts#L51](https://github.com/angular/universal/blob/master/modules/common/src/transfer_http.ts#L51)
and set cache to active again it works also for get requests done after a post request

```
if (req.method !== 'GET' && req.method !== 'HEAD') {
            this.isCacheActive = false;
            this.invalidateCacheEntry(req.url);
        } else {
            this.isCacheActive = true;
        }
```

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** by creating a github repo.

* **What is the expected behavior?**

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**

- Angular version: 5.0
- Browser: [Chrome 62]
- Language: [TypeScript 2.6 ]
- OS:  [Mac OS X  ]
- Platform: [NodeJs]

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, gitter, etc)

patrickmichalina:
Does your ApplicationRef.isStable ever return true?

lutzk:
yes i think so - i logged the value of `isStable` from:
`...node_modules/@nguniversal/common/src/transfer_http.js`

```
toPromise_1.toPromise
[...]

PR #984, opened 2018-05-10
Added caching for POST requests. Fixed bug for caching GET requests w…

vytautas-pranskunas-:
…ith params.

## PR Checklist

## PR Type
```
[X] Bugfix
[X] Feature
```

POST requests are not cached so flickering appears when using pure JSON API
Get requests cache does not respect request params

## What is the new behavior?
It should cache POST requests
Fixed bug when GET requests with different params are cached:
https://github.com/angular/universal/issues/927

```
[X] Yes
[ ] No
```

.forRoot() method should be called

## Other information

### DIFF
--- a/modules/common/src/transfer_http.ts
+++ b/modules/common/src/transfer_http.ts
@@ -6,100 +6,146 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {
-  HTTP_INTERCEPTORS,
-  HttpEvent,
-  HttpHandler,
-  HttpHeaders,
-  HttpInterceptor,
-  HttpRequest,
-  HttpResponse
+    HTTP_INTERCEPTORS,
+    HttpEvent,
+    HttpHandler,
+    HttpHeaders,
+    HttpInterceptor,
+    HttpRequest,
+    HttpResponse
 } from '@angular/common/http';
-import {ApplicationRef, Injectable, NgModule} from '@angular/core';
-import {BrowserTransferStateModule, TransferState, makeStateKey} from '@angular/platform-browser';
-import {Observable, of as observableOf} from 'rxjs';
-import {tap, take, filter} from 'rxjs/operators';
+import { ApplicationRef, Injectable, NgModule, InjectionToken, Inject, ModuleWithProviders } from '@angular/core';
+import { BrowserTransferStateModule, TransferState, makeStateKey, StateKey } from '@angular/platform-browser';
+import { Observable, of as observableOf } from 'rxjs';
+import { tap, take, filter } from 'rxjs/operators';
 
 export interface TransferHttpResponse {
-  body?: any | null;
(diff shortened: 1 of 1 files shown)


Toxicable:
Sorry but we're not going to merge the POST caching part.

However we'd be happy to put int the params part, that's a good feature to have put in.

vytautas-pranskunas-:
Why POST caching is not good? Because hen Ssr is only for REST api :(

patrickmichalina:
Generally POST requests are consider non-cacheable. It would be nice to allow a DI config value to support this though.

vytautas-pranskunas-:
Would it be ok if i add it to foroot call?

Toxicable:
@vytautas-pranskunas- No we cant take a forRoot sorry, it'll be too easy for users to accidentally enable.

While almost any of the HTTP verbs can be used however you like we prefer to stick to what they semantically mean. 
There's an issue here on the topic: https://github.com/angular/universal/issues/855

vytautas-pranskunas-:
I do not see how they can turn it by accident. i do not see any issue with cachable POST also because during ssr app_initializer, canActivate and similar calls are executed and those are not data changing requests they are like getters just in pure Json style. But probably you are right and we never know what can happen.  So i can add to forRoot call array of routes that should be cachable by POST. How does that sounds? Otherwise that's huge limitation of ssr because we cannot have it with pure Json api.

Toxicable:
@vytautas-pranskunas- That's my point. In your example your POST requests do not mutate the data, therefore they do not invalidate the already cached data.
Which is fine, as I mentioned above you can use the HTTP verbs however you like. 
But POST is intended to mutate data, not to retrieve it and that's the definition we're going to take on this.

We are happy to take the params part only, if you refactor it to just be that then I'll review it from there.

You're welcome to re-implement this in your own lib if you really need it that much, but for the general case we must remain consistent for everyone using it. Allowing POST to be cached would break a lot of apps without people knowing it.

vytautas-pranskunas-:
Hi,

should i make new pull request?
I have added to forRoot options obejct where you have to define POST request filter. SO by default POST caching is disabled but you can allow it and add different checks for example: allow caching for certain urls or any other. 
I think this would give pure JSOn api users ability to use SSR without braking anything accidenlty.

Toxicable:
A new pull request just for the Params caching would be good

vytautas-pranskunas-:
But i refactored whole POST request system. Or you saying that whole world should stic to REST Api if they want to use SSR? So you need to mention it before people pick framework. I think you are wrong and we all should look the best way to allow limitted POST caching.

EreckGordon:
What kind of post requests would be a good idea to cache? I'm drawing a blank.

vytautas-pranskunas-:
Non changing POST requests. Same as GET's. It is only a matter of protocol and approach people use. REST or JSON-Pure

PR #986, opened 2018-05-14
Added caching for POST requests.

vytautas-pranskunas-:
## PR Checklist

## PR Type

```
[X] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

Currently apps that using pure JSON api cannot use SSR because of lack of cache for POST requests

## What is the new behavior?
Intorduced controlled / limitted POST request caching

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/modules/common/src/transfer_http.ts
+++ b/modules/common/src/transfer_http.ts
@@ -6,100 +6,146 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {
-  HTTP_INTERCEPTORS,
-  HttpEvent,
-  HttpHandler,
-  HttpHeaders,
-  HttpInterceptor,
-  HttpRequest,
-  HttpResponse
+    HTTP_INTERCEPTORS,
+    HttpEvent,
+    HttpHandler,
+    HttpHeaders,
+    HttpInterceptor,
+    HttpRequest,
+    HttpResponse
 } from '@angular/common/http';
-import {ApplicationRef, Injectable, NgModule} from '@angular/core';
-import {BrowserTransferStateModule, TransferState, makeStateKey} from '@angular/platform-browser';
-import {Observable, of as observableOf} from 'rxjs';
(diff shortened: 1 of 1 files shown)


vytautas-pranskunas-:
Either we have to look for way to cache POST requests either people should know before choosing framework that with Angular they can only use REST api if they want to use SSR. I think my change offers highly controlled POST caching whih allows to use pure JSON api.

there are tones of reading about pure json api vs REST. 

> The move away from RESTful APIs has already begun. Both FaceBook’s Relay and Netflix’s Falcor dispense with any pretense of being RESTful.

https://mmikowski.github.io/json-pure/

Toxicable:
This will be my final word on this topic, I see no further reason to discuss it.

We wont be taking this, either as a opt-in or by default.
Take a look here: https://www.w3.org/Protocols/rfc2616/rfc2616-sec13.html
I quote from the HTTP spec 13.11:

>All methods that might be expected to cause modifications to the origin server's resources MUST be written through to the origin server. This currently includes all methods except for GET and HEAD. A cache MUST NOT reply to such a request from a client before having transmitted the request to the inbound server, and having received a corresponding response from the inbound server. This does not prevent a proxy cache from sending a 100 (Continue) response before the inbound server has sent its final reply.

---

As a work around for your specific case you can make an interceptor which will swap a GET into a POST. Therefore you can use `this.http.get` which will leaverage the caching, but then your intercrptor will swap it into the correct POST request.
This can be implemented in your application but it will not be apart of this libraray.

Also as I mentioned in your other PR we're happy to take the part on caching based on URL params, just not caching posts

PR #987, opened 2018-05-15
Respect GET params

vytautas-pranskunas-:
## PR Checklist

## PR Type

```
[X] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] angular.io application / infrastructure changes
[ ] Other... Please describe:
```

Cache is not respecting GET params

## What is the new behavior?
Cache is respecting GET params

```
[ ] Yes
[X] No
```

## Other information
No matter that my API is pure JSON and i am not using GET I fixed bug that GET requests where not cached by params. Just one smal remark - for the same reason that I am not using GET i was not able to test. I tested with POST so GET should work similar. Please test befor merge.

### DIFF
--- a/modules/common/src/transfer_http.ts
+++ b/modules/common/src/transfer_http.ts
@@ -6,100 +6,146 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {
-  HTTP_INTERCEPTORS,
-  HttpEvent,
-  HttpHandler,
-  HttpHeaders,
-  HttpInterceptor,
-  HttpRequest,
-  HttpResponse
+    HTTP_INTERCEPTORS,
+    HttpEvent,
+    HttpHandler,
+    HttpHeaders,
+    HttpInterceptor,
+    HttpRequest,
+    HttpResponse
 } from '@angular/common/http';
-import {ApplicationRef, Injectable, NgModule} from '@angular/core';
-import {BrowserTransferStateModule, TransferState, makeStateKey} from '@angular/platform-browser';
-import {Observable, of as observableOf} from 'rxjs';
-import {tap, take, filter} from 'rxjs/operators';
+import { ApplicationRef, Injectable, NgModule, InjectionToken, Inject, ModuleWithProviders } from '@angular/core';
+import { BrowserTransferStateModule, TransferState, makeStateKey, StateKey } from '@angular/platform-browser';
+import { Observable, of as observableOf } from 'rxjs';
+import { tap, take, filter } from 'rxjs/operators';
 
 export interface TransferHttpResponse {
-  body?: any | null;
-  headers?: {[k: string]: string[]};
-  status?: number;
-  statusText?: string;
-  url?: string;
+    body?: any | null;
+    headers?: { [k: string]: string[] };
+    status?: number;
+    statusText?: string;
+    url?: string;
 }
 
 function getHeadersMap(headers: HttpHeaders) {
-  const headersMap: {[name: string]: string[]} = {};
-  for (const key of headers.keys()) {
-    headersMap[key] = headers.getAll(key)!;
-  }
-  return headersMap;
+    const headersMap: { [name: string]: string[] } = {};
+    for (const key of headers.keys()) {
+        headersMap[key] = headers.getAll(key)!;
+    }
+    return headersMap;
 }
+
+export const TRANSFER_STATE_CACHE_KEY_TRANSFORMER = new InjectionToken('TRANSFER_STATE_CACHE_KEY_TRANSFORMER');
 
 @Injectable()
 export class TransferHttpCacheInterceptor implements HttpInterceptor {
 
-  private isCacheActive = true;
+    private isCacheActive = true;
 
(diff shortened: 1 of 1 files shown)


vytautas-pranskunas-:
Sry do not have time for it. What I done - 
1) joined params to object, stringify it and hashed. 
```
constructCacheKey method ...
```

2) Also i added check to not pass request through if ```transferState``` has value for it
```
if (!this.isCacheActive && !this.transferState.hasKey(storeKey)) {
            // Cache is no longer active. Pass the request through.
            return next.handle(req);
        }
```

3) Added invalidation after retrieving value from ```transferState```
```
if (this.transferState.hasKey(storeKey)) {
            // Request found in cache. Respond using it.
            const response = this.transferState.get(storeKey, {} as TransferHttpResponse);
            this.invalidateCacheEntry(storeKey);
...
```

modules/common/src/transfer_http.ts
```diff
@@ -6,111 +6,154 @@
...
-                url: event.url!,
-              });
-            }
-          })
-        );
+    private constructCacheKey(req: HttpRequest<any>) {
+        let key = `${req.method[0]}_${req.url}`;
+
+        key = key
+            .replace(/[^\w\s]/gi, '_');
```
Toxicable: Can you explain what this is doing?
A comment should suffice here

modules/common/src/transfer_http.ts
```diff
@@ -6,111 +6,154 @@
...
+                return { key: req.params.get(key) };
+            });
+
+            key += this.hashSeparator + this.hashParams(hashedParams);
+        }
+
+        return key;
+    }
+
+    private hashParams(body: any) {
```
Toxicable: What is this hashing method? 
Is it some well known method?
Why not use something off the shelf?

Toxicable:
@vytautas-pranskunas- If you cant resolve the above comments then we cannot accept this PR.

If that's ok with you then I'll implement the changes when I next get a chance

modules/common/src/transfer_http.ts
vytautas-pranskunas-: this just beutify key.

modules/common/src/transfer_http.ts
vytautas-pranskunas-: It is taken from stackowerlof. I thought why to take another dependency for such simple hashing functionality.

Toxicable: Can you add a comment to credit the author and source

modules/common/src/transfer_http.ts
```diff
@@ -6,111 +6,154 @@
...
+
+        key = key
+            .replace(/[^\w\s]/gi, '_');
+
+        if (req.method === 'GET') {
+            const hashedParams = req.params.keys().map((key: string) => {
+                return { key: req.params.get(key) };
+            });
+
+            key += this.hashSeparator + this.hashParams(hashedParams);
```
Toxicable: Doing it this way means if someone has the params in a different order it'll casue a new request.
EG
foo=2&bar=3  produces key of 25
bar=3&foo=2  produces key of 52

When since their content is the same they should produce the same key

vytautas-pranskunas-: you are right. But usually queries are not builded dynamically, they are fixed ant their order not changing. But ofcourse if you want - you can sort arasm before applying hashing.

modules/common/src/transfer_http.ts
vytautas-pranskunas-: https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery

vytautas-pranskunas-: https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery

Toxicable: Can you add that as a comment in your src code

modules/common/src/transfer_http.ts
wardbell: FWIW, param order should not matter. A reliable, case-sensitive (!) sort by key seems reasonable.

vytautas-pranskunas-: So you want to say that in one query it can be user=..&User=... and in another visa versa? Sounds like a smell...

Toxicable: The keys *should* be case senstive.
Meaning that the keys `user` and `User` are not equal.

vytautas-pranskunas-: I have added comment about sorting you asked

Toxicable: Sorry I didn't mean to make a comment about this, it needs to be applied aswell.
Merging this without the sorting will be broken.

Toxicable:
Closed in favour of https://github.com/angular/universal/pull/1005

PR #997, opened 2018-05-19
chore: remove Travis config

CaerusKaru:
We don’t build on Travis anymore, and so it’s not needed


CaerusKaru:
cc @vikerman to remove required status for Travis

PR #999, opened 2018-05-19
feat(socket-engine): introduce package

Toxicable:
## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit

## PR Type

```
[x] Feature
```


modules/socket-engine/README.md
```diff
@@ -0,0 +1,43 @@
...
+socketEngine.startSocketEngine(AppServerModuleNgFactory);
+```
+This will the socket engine which internally hosts a TCP Socket server.  
+The default port is `9090` and host of `localhost`
+You may want to leave this as a plain `.js` file since it is so simple and to make deploying it easier, but it can be easily transpiled from Typescript.  
+
+## Usage Client
+
+Your client can be whatever language, framework or platform you like.  
+Aslong as it can connect to a TCP Socket (which all frameworks can) then you're good to go.
```
vikerman: nit: As long

modules/socket-engine/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+This example will use JS for simplicity
+```typescript
+import * as net from 'net';
+
+const client = net.createConnection(9090, 'localhost', () => {
+  const renderOptions = {id: 1, url: '/path',
```
vikerman: This example makes it appear as if there is only one message sent per createConnection with id of 1.

Maybe separate out the connection from sending the request?

modules/socket-engine/src/main.ts
```diff
@@ -0,0 +1,55 @@
...
+  host = 'localhost',
+  port = 9090
+): Promise<SocketEngineServer> {
+  return new Promise((resolve, _reject) => {
+    const engine = new CommonEngine(moduleOrFactory);
+
+    const server = net.createServer(socket => {
+      socket.on('data', async buff => {
+        const message = buff.toString();
+        const renderOptions = JSON.parse(message) as SocketEngineRenderOptions;
```
bbortt: Hi all

I've been playing around with the socket-engine a bit and noticed an in my opinion significant error. My setup: Connect to the socket from an Apache Tomcat container. I am honestly new to sockets and followed [this guide](https://www.sitepoint.com/implementing-spring-websocket-server-and-client/), ended up something like this.

```
WebSocketClient simpleWebSocketClient = new StandardWebSocketClient();
List<Transport> transports = new ArrayList<>(1);
transports.add(new WebSocketTransport(simpleWebSocketClient));
SockJsClient sockJsClient = new SockJsClient(transports);

WebSocketStompClient stompClient = new WebSocketStompClient(sockJsClient);
stompClient.setMessageConverter(new MappingJackson2MessageConverter());

String url = "ws://localhost:9090/";
StompSessionHandler sessionHandler = new MyStompSessionHandler();
StompSession session = stompClient.connect(url, sessionHandler).get();
```

Now the problem is in the implementation of the Apache `org.apache.tomcat.websocket.WsWebSocketContainer.java` which sends some default parameters on the initial request. You can see the connect statement and it's unskippable parameters in [this file](https://github.com/apache/tomcat/blob/trunk/java/org/apache/tomcat/websocket/WsWebSocketContainer.java) on lines 266 to 275.

So here we are now, sending an initial request to the websocket. An informational request which is not in JSON format. This line tries to parse it - and dos not return an error -> The connection attempt ends in a timeout and the webapplication failes to start!

*As far as I see you should cover this line by a try-catch too and handle errors in a maner way?*

I'd like to state out once more that this is just what I noticed in a brief test. Also I'm not an Apache pro, just looked into the source code and didn't see a way to skip the initial request (except writing my own implementation, which is maybe a bit too overkill). Just wanted to let you know what I detected. It may be intentional that this line isn't covered by a try-catch, it may not. Up to you.

Cheers

bbortt: Oh that's the thing. Sorry and thank you for the explanation.

WORKSPACE
```diff
@@ -1,9 +1,9 @@
 workspace(name = "nguniversal")
 http_archive(
     name = "build_bazel_rules_nodejs",
-    url = "https://github.com/bazelbuild/rules_nodejs/archive/0.9.1.zip",
-    strip_prefix = "rules_nodejs-0.9.1",
-    sha256 = "6139762b62b37c1fd171d7f22aa39566cb7dc2916f0f801d505a9aaf118c117f",
+    # TODO: upgrade once https://github.com/bazelbuild/rules_nodejs/issues/218#issuecomment-395826361 is released
```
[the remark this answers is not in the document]
Toxicable: ran into some other issues with this change, will nbeed a full set of updates later

PR #1003, opened 2018-05-20
WIP: feat(grpc-engine): introduce package

Toxicable:
Similar implementation to https://github.com/angular/universal/pull/999
But even easier to implement on the other side due to the use of GRPC


Toxicable:
Blocked on https://github.com/bazelbuild/rules_typescript/issues/211

Twois:
Will it land in the near future? - And if will, can we get a more detailed explanation for?

Toxicable:
@Twois We're just blocked on a bazel issue, which we have a work around for now so will be able to progrss with it this weekend

alan-agius4:
Closing there doesn’t seem to be much interest in this from the community. 

Let’s reconsider this, if there is a demand for it.

duncandean:
This would definitely be useful to us. I’m going to take a closer look when I get back home and have reliable interwebz again. 🙂

Hopefully there is a greater interest in this.

KingDarBoja:
I was looking at the issue on the `ts_proto_library` as there is some `wrap` argument which is being set as `amd` by default, as it is not documented on the docs. Maybe that could give some insight on how to implement the **grpc-engine** for the _Framework and Platform agnostic Angular Universal rendering_ proposal.

Cheers!

PR #1004, opened 2018-05-20
refactor(common): reduce CommonEngine api

Toxicable:
- exports the private tokens from Common Engine
- exports less of the private api in Common Engine

### DIFF
--- a/modules/common/src/common-engine/engine.ts
+++ b/modules/common/src/common-engine/engine.ts
@@ -38,23 +38,23 @@ export class CommonEngine {
    * Render an HTML document for a specific URL with specified
    * render options
    */
-  render(filePath: string, opts: RenderOptions): Promise<string> {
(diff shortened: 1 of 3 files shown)


modules/common/src/common-engine/engine.ts
```diff
@@ -21,7 +21,7 @@ import {RenderOptions} from './interfaces';
 export class CommonEngine {
 
   /** Return an instance of the platformServer compiler */
-  getCompiler(): Compiler {
+  private getCompiler(): Compiler {
```
CaerusKaru: We're going to need this to be public for ASP.NET core when we switch over

Toxicable: Yup, so we should expose it then, I think keeping it as minimal as possible should be our goal with this

PR #1015, opened 2018-06-09
feat(hapi): upgrade to Hapi v17

### LINKED ISSUE #993: feat(hapi): update to Hapi v17
Our module currently only works with Hapi v16, but Hapi v17 has been out for a while and is not backwards-compatible.

CaerusKaru:
* fix RxJS globals
* add Hapi and Express versions to auto-stamping
* add unit tests
* update README with instructions

Fixes #993

### DIFF
--- a/integration/hapi-engine/package.json
+++ b/integration/hapi-engine/package.json
@@ -29,7 +29,7 @@
     "babel-loader": "^6.4.0",
     "babel-preset-es2015": "^6.22.0",
     "concurrently": "3.1.0",
-    "inert": "^4.2.1",
+    "inert": "^5.1.0",
     "protractor": "file:../../node_modules/protractor",
     "raw-loader": "^0.5.1",
     "webpack": "^2.2.1"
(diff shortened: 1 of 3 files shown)


modules/hapi-engine/README.md
```diff
@@ -18,20 +18,10 @@ server.connection({
...
 server.route({
   method: 'GET',
   path: '/{path*}',
-  handler: (req: Request, reply: Base_Reply) => {
-    hapiEngine({
-      req
-    }).then(html => reply(html))
-      .then(err => reply(boom.wrap(err)));
-  }
+  handler: (req: Request) => ngHapiEngine({req, bootstrap: ServerAppModule})
```
Toxicable: does it have to instantiate the whole engine per request? feels like a bit of overhead

CaerusKaru: Yeah, at least for now, because the argument for `ngHapiEngine` is `RenderOptions`, which has a mandatory `req`. And it's not instantiation. Remember: the engine is just a function wrapper around `renderPlatformServer`. It still does all the caching because that's handled in a constant external to the wrapper.

Toxicable:
Just needs a rebase

PR #1017, opened 2018-06-09
fix(common): fixup secondary entrypoints

Toxicable:
whops


Toxicable:
@CaerusKaru It still needs that for the actual source

PR #1051, opened 2018-08-15
WIP feat(express-engine): add schematics

CaerusKaru:
* Adds schematics to the Express Engine that allow
  for easier installation. This is basically an
  enhanced version of the vanilla Universal schematic
  that is available by default, where this one adds
  the Express Engine, a sample Express server, and
  the Webpack config, as well as some other necessary
  dependencies


modules/express-engine/schematics/README.md
```diff
@@ -0,0 +1,14 @@
+# Angular Universal Express-Engine Schematics
+A collection of Schematics for Angular Universal Express-Engine.
+
+## Collection
+
+### Install
+Adds Angular Universal Express Engine and its depedencies and pre-configures the application. 
+
+- Adds Express-Engine, NgModule-Factory-Loader, ts-loader, and webpack to `package.json`
```
CaerusKaru: Will finalize docs when behavior is finalized. Part of this was stolen from Material

modules/express-engine/schematics/install/index.spec.ts
```diff
@@ -0,0 +1,179 @@
+/**
```
CaerusKaru: More tests will be needed as behavior is finalized. These are the default tests, plus some added dependency checks. Will also need test runners that work with Bazel (Material has this already)

modules/express-engine/schematics/test-setup/bazel-test-init.ts
```diff
@@ -0,0 +1,43 @@
...
+/*
+ * NOTE: This file will run before the actual tests start inside of Bazel.
+ *
+ * It automatically runs before all spec files because the spec files are blocked
+ * until Jasmine runs the `describe`  blocks.
+ *
+ * We copy all needed files into the proper Bazel bin output in order to be able to test
+ * the schematics. Workaround for: https://github.com/bazelbuild/rules_typescript/issues/154
+ */
+
```
DevVersion: Maybe you could add a comment that refers to the original source of this file? This way it's clear that this file should be kept in sync with the file from Material.

modules/express-engine/schematics/test-setup/post-scheduled-tasks.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
CaerusKaru: This file can be removed. When it was taken from Material, they use it for their update task to run things sequentially. We don't, so we don't need it here.

modules/express-engine/schematics/install/index.ts
```diff
@@ -0,0 +1,142 @@
...
+    const pkgPath = '/package.json';
+    const buffer = host.read(pkgPath);
+    if (buffer === null) {
+      throw new SchematicsException('Could not find package.json');
+    }
+
+    const pkg = JSON.parse(buffer.toString());
+
+    pkg.dependencies['@nguniversal/express-engine'] = '0.0.0-PLACEHOLDER';
+    pkg.dependencies['@nguniversal/module-map-ngfactory-loader'] = '6.0.0' || '0.0.0-PLACEHOLDER';
```
vikerman: Why is the "||" needed? Isn't it a no-op?

PR #1059, opened 2018-08-24
feat(express-engine): add option to skip Universal schematic

CaerusKaru:


modules/express-engine/schematics/install/index.ts
```diff
@@ -87,6 +87,11 @@ function updateConfigFile(options: UniversalOptions): Rule {
         ]
       }
     };
+
+    if (!clientProject.architect.server) {
```
vikerman: I don't understand what this is for - Can you add a comment?

CaerusKaru: Sure, it's there because if the Universal schematic wasn't run, there isn't necessarily a server entry in the project config

vikerman: ok - Please add a comment in the code. Thanks

CaerusKaru: Done

PR #1081, opened 2018-10-11
fix(express-engine): add webpack option to schematics

CaerusKaru:
* Enable webpack option by default
* Bump karma to version 3 to fix audit issues

### DIFF
--- a/modules/express-engine/schematics/install/index.ts
+++ b/modules/express-engine/schematics/install/index.ts
@@ -7,22 +7,23 @@
  */
 import {experimental, JsonObject, strings} from '@angular-devkit/core';
 import {
+  apply,
+  chain,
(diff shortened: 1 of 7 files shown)


modules/express-engine/schematics/install/index.ts
```diff
@@ -48,16 +81,13 @@ function addDependenciesAndScripts(options: UniversalOptions): Rule {
 
     const pkg = JSON.parse(buffer.toString());
 
-    pkg.dependencies['@nguniversal/express-engine'] = '0.0.0-PLACEHOLDER';
```
vikerman: Why were @nguniversal/express-engine and module-map-ngfactory-loader removed?

PR #1083, opened 2018-10-13
WIP: build: update Bazel to use fine-grained deps

CaerusKaru:


CaerusKaru:
Before I forget, credit for a lot of this goes to @DevVersion's incredible work over on Material.

CaerusKaru:
This is blocked on the `test` step because `socket-engine`, `hapi-engine`, and `express-engine` tests bring in `platform-server`, which doesn't wire up correctly with `domino` and `xhr2`.

integration/common/package.json
```diff
@@ -8,15 +8,15 @@
     "url": "https://github.com/angular/angular.git"
   },
   "dependencies": {
-    "@angular/animations": "file:../../node_modules/@angular/animations",
-    "@angular/common": "file:../../node_modules/@angular/common",
+    "@angular/animations": "^7.1.4",
+    "@angular/common": "^7.1.4",
     "@angular/compiler": "file:../../node_modules/@angular/compiler",
```
Toxicable: why are some of these files some version?

CaerusKaru: They're no longer installed at the top-level because Angular is now built from source with Bazel

Toxicable: but there's still a few that are manually installed?

PR #1091, opened 2018-11-07
fix(common): Update transfer_http cache to not invalidate after mutating calls

JediSange:
Ran into this issue while working in Angular Universal 6.0. We noticed that the transfer state wasn't accurately returning and that our page, despite having the initial transfer state rendered on the page, was still making an HTTP call to fetch the data because it was producing a cache miss here.

This was happening because we were submitting PUT/POST requests for reporting at a an earlier point in time. In my opinion, there is not a good reason to invalidate this cache for any reason other than the application becoming stable. This code reflects that and it was a simple enough change for our usage so we are trying to commit it back.

### DIFF
--- a/modules/common/src/transfer_http.ts
+++ b/modules/common/src/transfer_http.ts
@@ -70,10 +70,10 @@ export class TransferHttpCacheInterceptor implements HttpInterceptor {
   }
 
   intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
-    // Stop using the cache if there is a mutating call.
+    // Skip the cache entirely for calls that seek to mutate data.
     if (req.method !== 'GET' && req.method !== 'HEAD') {
-      this.isCacheActive = false;
       this.invalidateCacheEntry(req.url);
+      return next.handle(req);
     }
 
     if (!this.isCacheActive) {


vikerman:
Hi - This would be a breaking change for whoever is using this library.

This is relatively small piece of code meant more or less to be a reference implementation. If you wanted a different behvior you can just fork this piece of code.

Thanks.

Toxicable:
You can read the discussions here on why we don't cache mutating http methods
https://github.com/angular/universal/pull/984
https://github.com/angular/universal/pull/986

JediSange:
@vikerman In what ways would this be a breaking change? It wouldn't affect any GET or HEAD requests. You arbitrarily deem that an application should not want any of the cached data from the `transferState` after receiving a PUT/POST/etc. Perhaps that decision is not arbitrary and I'd love to hear more about why that is -- because that doesn't seem like a common paradigm. If I am showing some record, I still want to reference that pre-fetched data from the transferState even if I have also fired reporting calls.

@Toxicable My change does not cache mutating http calls. This bypasses caching them, but also doesn't invalidate the existing GET/HEAD requests the server has already done for us.

Toxicable:
@JediSange You're still reading from the cache in subsequence non mutating calls after a mutationing request. That's the issue.
Once a mutation is made the cache is invalid from that point in time.
We don't know what your specific mutating calls actually effect, therefore the only general sulution is to invalidate the whole cache and never read from it again.

JediSange:
That's a fair perspective on it and I appreciate the explanation. However, there is currently no way to specify a request as mutating for the application state versus just reporting data, as is in my case. It's very common to have PUT/POST requests fired from the application for any myriad of reasons that do not lead to mutating the application state.

Perhaps a solution for me is having a second instance of the HttpClient service that only handles reporting, and not have the interceptor on that instance of the service. Either way, I'd be curious to hear feedback about this as I still see open requests related to screen flickering with TransferState. It took us a really long time to debug this as it's not documented anywhere about the HttpTransferState as well. Would a PR for updating that documentation be preferred?

PR #1102, opened 2018-12-07
cache nested get calls between post calls

zovi-wynk:
nested get api calls are not cached if post calls is present in between.


CaerusKaru:
Duplicate of #1091

PR #1108, opened 2018-12-21
feat(builders): introduce ssr builder

### LINKED ISSUE #1022: Rebuild angular universal on save
I turned my application to angular universal. It's great but I have a problem with developing mode. I need see my editing result real-time. I need restart my app on each atrl+s (save) like `"build --watch"` and `"nodmeon"`.

My server is expressjs and I'm using `"npm run build:ssr"` after each edit. It takes lots of time!!

How can I do that?

Toxicable:
Fixes #1022


Toxicable:
Unfortunately we've hit a blocker with this for the time being.
While in it's current state it can build all the targets in watch mode and parallel, which greatly improves the workflow for devs working on Universal.

However this has 2 issues with it.
- The initial compile will fail since the front wont be built yet, but the nodejs app depends on it's output.
- There is no "live reload" as has been expected.

So until we have solved those issues we're going to hold off on this. The main blocker is some new APIs from the CLI which should drop around 8.0, meaning you can expect a beta release around then.

vikerman:
We should use the new architect API. Closing this for now.

PR #1111, opened 2018-12-25
fix(readme): use circle ci build badge, update img url, add socketengine

Toxicable:


Toxicable:
see https://github.com/angular/universal/tree/fabian/readme-updates

PR #1112, opened 2018-12-25
test: add tests to ensure that document.location is populated

Toxicable:
small test to ensure that window.location is populated by what we provide to the engine which should be populated by platform-server


Toxicable:
Ah yes
> ERROR ReferenceError: window is not defined

Meaning that platformServer isn't populating window.location!?

PR #1130, opened 2019-02-19
feat(express-engine): add ModuleMapLoaderModule to the app server imports during ng-add

### LINKED ISSUE #1129: ng add @nguniversal/express-engine should add ModuleMapLoaderModule to App server module
# Bug Report

#### What is the expected behavior?
`ng add @nguniversal/express-engine` should support lazy loaded routes as well.

#### What is the current behavior?
Exception is thrown on the server when rendering lazy loaded routes until ModuleMapLoaderModule is added to the AppServerModule.

#### What modules are related to this issue?
```
- [x] aspnetcore-engine
- [ ] common
- [ ] express-engine
- [x] hapi-engine
```

#### Environment:
**@nguniversal versions**
  - aspnetcore-engine: 
  - common:
  - express-engine:
  - hapi-engine:
  - module-map-ngfactory-loader:
```
<!--
Output from: `ng --version`.
If nothing, output from: `node --version` and `npm --version`.
  Windows (7/8/10). Linux (incl. distribution). macOS (El Capitan? Sierra?)
-->
```

#### Is there anything else we should know?

CaerusKaru:
Fixes #1129

### DIFF
--- a/modules/express-engine/schematics/install/index.ts
+++ b/modules/express-engine/schematics/install/index.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {experimental, strings} from '@angular-devkit/core';
+import {experimental, strings, normalize} from '@angular-devkit/core';
 import {
   apply,
   chain,
@@ -28,6 +28,15 @@ import {
   NodeDependencyType,
 } from '@schematics/angular/utility/dependencies';
 import {BrowserBuilderOptions} from '@schematics/angular/utility/workspace-models';
+import {getProject} from '@schematics/angular/utility/project';
+import {
+  getProjectTargets,
+  targetBuildNotFoundError,
+} from '@schematics/angular/utility/project-targets';
+import {InsertChange} from '@schematics/angular/utility/change';
+import {addSymbolToNgModuleMetadata, insertImport} from '@schematics/angular/utility/ast-utils';
+import * as ts from 'typescript';
+import {findAppServerModulePath} from './utils';
 
 // TODO(CaerusKaru): make these configurable
 const BROWSER_DIST = 'dist/browser';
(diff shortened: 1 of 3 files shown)


modules/express-engine/schematics/install/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+  return exportDeclaration;
+}
+
+export function findAppServerModulePath(host: Tree, mainPath: string): string {
+  const exportDeclaration = findAppServerModuleExport(host, mainPath);
+  if (!exportDeclaration) {
+    throw new SchematicsException('Export declaration not found');
```
vikerman: It's not clear to me from the error message what 'Export declaration' is referring to. Maybe specify Internal Error if the error is not actually actionable by the user?

modules/express-engine/schematics/install/index.ts
```diff
@@ -141,6 +150,59 @@ function updateConfigFile(options: UniversalOptions): Rule {
...
+    const mainPath = normalize('/' + clientTargets.server.options.main);
+
+    const appServerModuleRelativePath = findAppServerModulePath(host, mainPath);
+    const modulePath = normalize(
+      `/${clientProject.root}/src/${appServerModuleRelativePath}.ts`);
+
+    // Add the module map loader import
+    let moduleSource = getTsSourceFile(host, modulePath);
+    const importModule = 'ModuleMapLoaderModule';
+    const importPath = '@nguniversal/module-map-ngfactory-loader';
```
vikerman: Is the package.json already including @nguniversal/module-map-ngfactory-loader ?

CaerusKaru: Yep!

modules/express-engine/schematics/install/utils.ts
CaerusKaru: It's referring to the `export {DECLARATION}` block. Do you have a suggestion to make that more apparent?

vikerman: I would have the error message indicate what sort of export declaration it was expecting - Like "Exported AppServer Module not found" or something like that instead of a generic statement like the current one.

PR #1167, opened 2019-05-24
fix(express-engine): add bundleDependencies and lazy-loading fixes

CaerusKaru:


vikerman:
Fixes https://github.com/angular/angular-cli/issues/12273 !!

vikerman:
Fixes https://github.com/angular/angular-cli/issues/8616

MarkPieszak:
Wow !! I didn't know you guys were working on this - do you need me to look into anything / test it with some aspnet core apps? Amazing 👏

vikerman:
@MarkPieszak - We are essentially moving anything @angular or @nguniversal related to main.server.ts so that we don't end up with two copies of Angular.

How does this affect ast.net setup?

PR #1206, opened 2019-07-19
feat(schematics): add prerendering scripts to express-schematic

### LINKED ISSUE #1201: Static site pre-rendering entrypoint
We should consider adding the prerender script as part of the schematics so that Universal apps can be build time prerendered.

This needs a design around how the URLs to be prerendered are configured/discovered.

MarkPieszak:
Fixes #1201

- Prerendering Added
- Sitemap auto creation for prerendering
  - Has "hostname" where you prepend your URL for all routes (https://mysite.com) etc
- Updated webpack, tsconfig conditionally if prerendering is skipped
- Basic prerender script setup
  - Errors in prerendering get propagated so if anything goes wrong it won't finish.

routes.json
Routes create sitemap.xml for prerendering.

Example:
image


modules/express-engine/schematics/install/files/root/__prerenderFileName@stripTsExtension__.ts
```diff
@@ -0,0 +1,46 @@
...
+import 'reflect-metadata';
+import {readFileSync, writeFileSync, existsSync, mkdirSync} from 'fs';
+import {join} from 'path';
+
+import {enableProdMode} from '@angular/core';
+// Faster server renders w/ Prod mode (dev mode never needed)
+enableProdMode();
+
+// Import module map for lazy loading
+import {provideModuleMap} from '@nguniversal/module-map-ngfactory-loader';
```
vikerman: We are now exporting all Anghuar related symbols from `main.server.ts` to avoid bundling multiple copies of Angular. Can this script also import all required symbols from `main.server.ts` just like we do with `server.ts` ?

MarkPieszak: Done! :+1:

modules/express-engine/schematics/install/files/root/__prerenderFileName@stripTsExtension__.ts
```diff
@@ -0,0 +1,43 @@
+import 'zone.js/dist/zone-node';
+
+import 'reflect-metadata';
+import {readFileSync, writeFileSync, existsSync, mkdirSync} from 'fs';
+import {join} from 'path';
+
+import {enableProdMode} from '@angular/core';
```
vikerman: Remove `enableProdMode()` and the import. This is already done in `main.server.ts`

MarkPieszak: Fixed!

vikerman:
Please see if you can add an option to the schematics to just add the prender parts - So that  it can be added to an existing Universal project.

MarkPieszak:
Added Flag `--updatePrerenderOnly` (defaults to False), to _only_ add prerender scripts, file, and add the needed exports to `main.server`

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
```diff
@@ -16,5 +16,7 @@
       "dom"
     ]
   },
-  "include": ["<%= stripTsExtension(serverFileName) %>.ts"]
+  "include": [
+    "<%= stripTsExtension(serverFileName) %>.ts",
+    "<%= stripTsExtension(prerenderFileName) %>.ts"
```
alan-agius4: Shouldn't this be excluded when using `skipPrerender`?

modules/express-engine/schematics/install/files/root/webpack.server.config.js
```diff
@@ -7,7 +7,9 @@ module.exports = {
   mode: 'none',
   entry: {
     // This is our Express server for Dynamic universal
-    server: './<%= stripTsExtension(serverFileName) %>.ts'
+    server: './<%= stripTsExtension(serverFileName) %>.ts',
+    // This is our script for Static Prerendering
+    prerender: './<%= stripTsExtension(prerenderFileName) %>.ts'
```
alan-agius4: Shouldn't this be excluded when using `skipPrerender`?

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
alan-agius4: I cannot find anywhere were an existing the `server.tsconfig` is being modified, since when not using `skipUniversal` option the tsconfig will be created by the `@schematics/angular` universal schematic.

https://github.com/angular/universal/blob/2aeb2ba67cf0043bc4b1620e855d77e33f14b83e/modules/express-engine/schematics/install/index.ts#L292

https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/universal/files/root/__tsconfigFileName__.json.template

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
```diff
@@ -16,5 +16,7 @@
       "dom"
     ]
   },
-  "include": ["<%= stripTsExtension(serverFileName) %>.ts"]
+  "include": [
```
alan-agius4: You are missing the closing `]`

modules/express-engine/schematics/install/files/root/__prerenderFileName@stripTsExtension__.ts
```diff
@@ -0,0 +1,42 @@
...
+ROUTES.forEach(route => {
+  const fullPath = join(BROWSER_FOLDER, route);
+
+  // Make sure the directory structure is there
+  if (!existsSync(fullPath)) {
+    mkdirSync(fullPath);
+  }
+
+  // Writes rendered HTML to index.html, replacing the file if it already exists.
+  previousRender = previousRender.then(_ => renderModuleFactory(AppServerModuleNgFactory, {
```
alan-agius4: If there is an error here, would this throw with unhandled promise rejection?

modules/express-engine/schematics/install/files/root/__prerenderFileName@stripTsExtension__.ts
```diff
@@ -0,0 +1,42 @@
+import 'zone.js/dist/zone-node';
+
+import 'reflect-metadata';
+import {readFileSync, writeFileSync, existsSync, mkdirSync} from 'fs';
+import {join} from 'path';
+
+// * NOTE :: leave this as require() since this file is built Dynamically from webpack
+const {AppServerModuleNgFactory, LAZY_MODULE_MAP, provideModuleMap, renderModuleFactory, enableProdMode} = require('./<%= getServerDistDirectory() %>/main');
```
alan-agius4: Aren't `renderModuleFactory`, `enableProdMode` in `@angular/platform-server` and `@angular/core`?

modules/express-engine/schematics/install/files/root/__prerenderFileName@stripTsExtension__.ts
MarkPieszak: Yes it actually throws it regardless, for example a `(<any>window).test = '123';` thrown in an app components code (that will pass ng build, etc) will throw an error when trying to build the prerender like shown below.

image

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
MarkPieszak: Thank you, fixed!

modules/express-engine/schematics/install/files/root/webpack.server.config.js
MarkPieszak: Yes it should, let me know if I did the templating incorrectly, I couldn't find many good examples within other CLI schematics! :)

```ts
<%= skipPrerender ? '' : '"<%= stripTsExtension(prerenderFileName) %>.ts"' %>
```

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
```diff
@@ -16,5 +16,8 @@
...
+    "<%= stripTsExtension(serverFileName) %>.ts"<%= skipPrerender ? '' : ',' %>
```
alan-agius4: This can be simplified a bit.

```
  "include": [
    "<%= stripTsExtension(serverFileName) %>.ts"<% if (!skipPrerender) { %>,
     <%= stripTsExtension(prerenderFileName) %>.ts"<% } %>
  ]
```

modules/express-engine/schematics/install/files/root/webpack.server.config.js
```diff
@@ -7,7 +7,10 @@ module.exports = {
   mode: 'none',
   entry: {
     // This is our Express server for Dynamic universal
-    server: './<%= stripTsExtension(serverFileName) %>.ts'
+    server: './<%= stripTsExtension(serverFileName) %>.ts',
+    <%= skipPrerender ? "" : "// This is our script for Static Prerendering" %>
+    <%= skipPrerender ? "" : "prerender: './<%= stripTsExtension(prerenderFileName) %>.ts'"
+
```
alan-agius4: Nit: remove extra line

modules/express-engine/schematics/install/files/root/webpack.server.config.js
```diff
@@ -7,7 +7,10 @@ module.exports = {
   mode: 'none',
   entry: {
     // This is our Express server for Dynamic universal
-    server: './<%= stripTsExtension(serverFileName) %>.ts'
+    server: './<%= stripTsExtension(serverFileName) %>.ts',
+    <%= skipPrerender ? "" : "// This is our script for Static Prerendering" %>
```
alan-agius4: Same here

```
  entry: {
    // This is our Express server for Dynamic universal
    server: './<%= stripTsExtension(serverFileName) %>.ts'<% if(!skipPrerender) { %>,
    // This is our script for Static Prerendering
    "prerender": './<%= stripTsExtension(prerenderFileName) %>.ts' %><% } %>
  },
```

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
MarkPieszak: You can do if's 😆 🤦‍♂ 
I was working on this way way too early this morning!
Let me clean this up

alan-agius4: Haha happens to me as well 😜

modules/express-engine/schematics/install/files/root/webpack.server.config.js
MarkPieszak: Fixed!

modules/express-engine/schematics/install/files/root/webpack.server.config.js
MarkPieszak: Fixed!

modules/express-engine/schematics/install/files/root/__serverFileName@stripTsExtension__.tsconfig.json
MarkPieszak: Fixed everything thanks again Alan, great catches :)

modules/express-engine/schematics/install/index.ts
```diff
@@ -194,12 +219,16 @@ function addExports(options: UniversalOptions): Rule {
     const mainSourceFile = getTsSourceFile(host, mainPath);
     let mainText = getTsSourceText(host, mainPath);
     const mainRecorder = host.beginUpdate(mainPath);
+    const enableProdModeExport = generateExport(mainSourceFile, ['enableProdMode'],
```
alan-agius4: Nit: we should actually check if these exports already exists prior to adding them. Since, we will add these exports in the universal schematics as well

modules/express-engine/schematics/install/index.ts
```diff
@@ -194,12 +219,16 @@ function addExports(options: UniversalOptions): Rule {
...
+      '@angular/core');
+    const renderModuleFactoryExport = generateExport(mainSourceFile, ['renderModuleFactory'],
```
alan-agius4: Will this target version 8 or 9? Is so probable we should change `renderModuleFactory` to `renderModule`

Some more context here: https://github.com/angular/angular-cli/pull/15517

alan-agius4: Actually regarding the later two comments if we wait for version 9 you don’t need to add them because they will be adde by default in the cli version 9.

Also, for existing project I will do a migration to add these exports

alan-agius4: Actually regarding the later two comments if we wait for version 9 you don’t need to add them because they will be adde by default in the cli version 9.

Also, for existing project I will do a migration to add these exports

modules/express-engine/schematics/install/files/root/__prerenderFileName@stripTsExtension__.ts
```diff
@@ -0,0 +1,53 @@
...
+import {join} from 'path';
+import * as fs from 'fs';
+
+// * NOTE :: leave this as require() since this file is built Dynamically from webpack
+const {AppServerModuleNgFactory, LAZY_MODULE_MAP, provideModuleMap, renderModuleFactory, enableProdMode} = require('./<%= getServerDistDirectory() %>/main');
+
+const routeData = require('./routes.json');
+
+// Faster server renders w/ Prod mode (dev mode never needed)
+enableProdMode();
```
alan-agius4: Is this still needed since it's already defined in main.server.ts?

modules/express-engine/schematics/install/schema.json
```diff
@@ -25,6 +25,11 @@
       "format": "path",
```
alan-agius4: `skipPrerender` doesn't seem to be defined in the schema.

vikerman:
Superseded by #1357

PR #1229, opened 2019-09-06
fix(express-engine): add missing `@types/express` for express

alan-agius4:


alan-agius4:
Related to https://github.com/angular/angular-cli/pull/14905

PR #1234, opened 2019-09-09
build: update angular dependencies

alan-agius4:
1) With this change we update to the latest pre-release angular dependencies.
2) We also disable Ivy compiler for the current tests
3) Enable dts bundling for universal packages


modules/aspnetcore-engine/tokens/public_api.ts
```diff
@@ -6,4 +6,4 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-export * from './tokens';
+export * from './injection-tokens';
```
alan-agius4: This is needed because of https://github.com/angular/universal/pull/1234/commits/2ee137ff47f840fde483b5d22d6a6b290b1cfc29

alan-agius4: Already reported in https://github.com/angular/angular/issues/20931

PR #1247, opened 2019-09-21
fix(express-engine): export Express app from server.ts

vikerman:
Change the schematics to export the Express app object itself from
server.ts.

This can be used from a serverless function to call into the Express app
directly. Example -
https://github.com/angular/angularfire2/blob/master/docs/universal/cloud-functions.md

### DIFF
--- a/integration/express-engine-ivy/server.ts
+++ b/integration/express-engine-ivy/server.ts
@@ -6,33 +6,34 @@ import { join } from 'path';
 
 import { AppServerModule } from './src/main.server';
 
+// The Express app is exported so that it can be used by serverless Functions.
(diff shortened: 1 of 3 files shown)


vikerman:
@alan-agius4 ^^^

alan-agius4:
I guess we need to do that same thing for hapi-engine, right?

vikerman:
Didn't do it for hapi-engine because it wasn't clear whether server less solutions work with hapi the same way they do with Express.

But looks like there might be adapters for Hapi also - 
https://www.npmjs.com/package/serverless-hapi

Will send similar PR for Hapi.

PR #1248, opened 2019-09-21
feat(common): set RouterModule initialNavigation in schematics

### LINKED ISSUE #1200: Add `initialNavigation: 'enabled'` to schematics
We have a lot of issues reported with  flickering in lazy routes in Universal. The solution is to set `initialNavigation: 'enabled'`.

This is a good default for Universal apps and we should automatically try to change `app.routing.module.ts` to set this flag as part of `ng add @nguniversal/express-engine`

CaerusKaru:
Fixes #1200


CaerusKaru:
Note: still needs a unit test. Will consult with @alan-agius4

modules/common/schematics/install/index.ts
```diff
@@ -120,6 +124,135 @@ function updateServerTsConfigRule(options: UniversalOptions): Rule {
...
+    const basePath = process.cwd();
+    const {config} = ts.readConfigFile(tsConfigPath, ts.sys.readFile);
+    const parseConfigHost = {
+      useCaseSensitiveFileNames: ts.sys.useCaseSensitiveFileNames,
+      fileExists: ts.sys.fileExists,
+      readDirectory: ts.sys.readDirectory,
+      readFile: ts.sys.readFile,
+    };
+    const parsed = ts.parseJsonConfigFileContent(config, parseConfigHost, basePath, {});
+    const tsHost = ts.createCompilerHost(parsed.options, true);
```
alan-agius4: Using a tsProgram and compilerHost is a schematic is not quite the norm. 

When using a compiler host at the very minimum you should override the readFile method for two main reasons.
1) Schematics uses a virtual file system and the contents of a file on disk might not be the latest version. 
2) You need to handle BOM replacement See: https://github.com/angular/angular/pull/30719

Ex
```ts
      host.readFile = fileName => {
        const buffer = tree.read(relative(basePath, fileName));
        // Strip BOM as otherwise TSC methods (Ex: getWidth) will return an offset which
        // which breaks the CLI UpdateRecorder.
        // See: https://github.com/angular/angular/pull/30719
        return buffer ? buffer.toString().replace(/^\uFEFF/, '') : undefined;
      };
```

alan-agius4: Using a tsProgram and compilerHost is a schematic is not quite the norm. 

When using a compiler host at the very minimum you should override the readFile method for two main reasons.
1) Schematics uses a virtual file system and the contents of a file on disk might not be the latest version. 
2) You need to handle BOM replacement See: https://github.com/angular/angular/pull/30719
3) Type checking is slower

Ex
```ts
      host.readFile = fileName => {
        const buffer = tree.read(relative(basePath, fileName));
        // Strip BOM as otherwise TSC methods (Ex: getWidth) will return an offset which
        // which breaks the CLI UpdateRecorder.
        // See: https://github.com/angular/angular/pull/30719
        return buffer ? buffer.toString().replace(/^\uFEFF/, '') : undefined;
      };
``` 

An alternative to find the routing module is to use one of the internal helpers from @schematics/angular. To be more precise `findBootstrapModuleCall` and `findBootstrapModulePath`. An example of the usage can be seen here https://github.com/angular/angular-cli/blob/master/packages/schematics/angular/universal/index.ts

modules/common/schematics/install/index.ts
```diff
@@ -18,9 +18,13 @@ import {
   findPropertyInAstObject,
```
alan-agius4: For UT, you can do something similar to 
https://github.com/angular/angular-cli/blob/703f74e72a04f45075f6d3a113912ae682346048/packages/schematics/angular/universal/index_spec.ts#L73-L80

modules/common/schematics/install/index.ts
```diff
@@ -18,9 +18,13 @@ import {
...
   appendValueInAstArray,
 } from '@schematics/angular/utility/json-utils';
+import {relative} from 'path';
```
alan-agius4: You should use the devkit path helpers

CaerusKaru:
Blocked on https://github.com/angular/angular/pull/33128

modules/common/schematics/install/index.ts
```diff
@@ -120,6 +124,135 @@ function updateServerTsConfigRule(options: UniversalOptions): Rule {
...
+
+        ts.forEachChild(node, visitNode);
+      });
+
+      if (routerModuleNode) {
+        const update = host.beginUpdate(relative(basePath, sourceFile.fileName));
+        update.remove(routerModuleNode.getStart(), routerModuleNode.getWidth());
+        update.insertRight(
+          routerModuleNode.getStart(),
+          printer.printNode(
```
alan-agius4: If you are using `printNode` I strongly suggest to use `singleQuote` private option. This is because it will break some formatting otherwise, because the CLI projects are typically using single quotes but TypeScript uses double by default.

Ex
```
      // TypeScript will emit the Node with double quotes.
      // In schematics we usually write code with a single quotes
      // tslint:disable-next-line: no-any
      (moduleSpecifier as any).singleQuote = true;

      const newExportDeclarationText = printer.printNode(
        ts.EmitHint.Unspecified,
        ts.createExportDeclaration(
          undefined,
          undefined,
          ts.createNamedExports(exportSpecifiers),
          moduleSpecifier,
        ),
        source,
      );
```

modules/common/schematics/install/index.ts
```diff
@@ -120,6 +124,135 @@ function updateServerTsConfigRule(options: UniversalOptions): Rule {
...
+    const {config} = ts.readConfigFile(tsConfigPath, ts.sys.readFile);
+    const parseConfigHost = {
+      useCaseSensitiveFileNames: ts.sys.useCaseSensitiveFileNames,
+      fileExists: ts.sys.fileExists,
+      readDirectory: ts.sys.readDirectory,
+      readFile: ts.sys.readFile,
+    };
+    const parsed = ts.parseJsonConfigFileContent(config, parseConfigHost, basePath, {});
+    const tsHost = ts.createCompilerHost(parsed.options, true);
+    const program = ts.createProgram(parsed.fileNames, parsed.options, tsHost);
```
alan-agius4: Thinking about this, I don't think we should create a program. 

We can use the approach used in lazy-loading migrations which is to visit all the source files without having to create a full type-script program, https://github.com/angular/angular-cli/blob/7df180d9c9dd3540811b6bd634772c08b774a382/packages/schematics/angular/migrations/update-8/update-lazy-module-paths.ts and `getRouterModuleDeclaration` in https://github.com/angular/angular-cli/blob/7df180d9c9dd3540811b6bd634772c08b774a382/packages/schematics/angular/utility/ast-utils.ts#L579 method would come in handle.

Splaktar:
I think that the current plan is for https://github.com/angular/angular/pull/33128 to be pushed to v10?

In that case, should this be moved out of the 9.0 milestone?

CaerusKaru:
@Splaktar This will still be part of v9. Once the model in core has been changed, this schematic will likewise be updated but will still be needed.

modules/common/schematics/add/index.spec.ts
```diff
@@ -18,18 +21,34 @@ describe('Add Schematic Rule', () => {
...
+  let runner: SchematicTestRunner;
+  let host: TempScopedNodeJsSyncHost;
+  let tree: UnitTestTree;
+  let tmpDirPath: string;
+  let previousWorkingDir: string;
 
   beforeEach(async () => {
-    appTree = await createTestApp();
-    schematicRunner = new SchematicTestRunner('schematics', collectionPath);
+    runner = new SchematicTestRunner('schematics', collectionPath);
```
alan-agius4: I think the below can be reverted no?

modules/common/schematics/add/index.ts
```diff
@@ -169,6 +180,88 @@ function updateServerTsConfigRule(options: AddUniversalOptions): Rule {
...
+      return;
+    }
+
+    const tsConfigPath = serverTarget.options.tsConfig;
+    if (!tsConfigPath || typeof tsConfigPath !== 'string' || !host.exists(tsConfigPath)) {
+      // No tsconfig path
+      return;
+    }
+
+    const basePath = process.cwd();
```
alan-agius4: Basepath shouldn’t be process.cwd but rather workspace root.

This is because the ng add schematic can be executed from a project directory example (projects/my-app) and the host paths are also relative to the workspace.

alan-agius4:
@googlebot I consent.

modules/common/schematics/add/index.ts
```diff
@@ -169,6 +179,119 @@ function updateServerTsConfigRule(options: AddUniversalOptions): Rule {
...
+    };
+    tsHost.getCurrentDirectory = function () {
+      return host.root.path;
+    };
+    tsHost.resolveModuleNames = function (
+      moduleNames: string[],
+      containingFile: string,
+    ): (ts.ResolvedModule | undefined)[] {
+      return moduleNames.map(moduleName => {
+        if (moduleName.startsWith('.')) {
```
alan-agius4: I think path mappings will not be supported with this. We need to check what happens

modules/common/schematics/add/index.spec.ts
```diff
@@ -92,4 +95,11 @@ describe('Add Schematic Rule', () => {
       'server.ts',
     ]);
   });
+
+  it(`should set 'initialNavigation' to enabled`, async () => {
```
alan-agius4: Note: Add a test to verify that properties are appended to existing object literals.

modules/common/schematics/add/index.spec.ts
```diff
@@ -92,4 +95,11 @@ describe('Add Schematic Rule', () => {
...
+    const routerPath = '/projects/test-app/src/app/app-routing.module.ts';
+    const tree = await schematicRunner
+      .callRule(addUniversalCommonRule(defaultOptions), appTree).toPromise();
+    expect(tree.read(routerPath)!.toString()).toContain(`initialNavigation: 'enabled'`);
```
alan-agius4: We should probably improve the expect here

Something like;
```
    expect(tree.read(routerPath)!.toString()).toContain(`forRoot(routes, { initialNavigation: 'enabled' }`);
```

PR #1256, opened 2019-09-25
Builders package skeleton

alan-agius4:
Note: there is no testing setup yet. But this should be enough to get started

//cc @manfredsteyer

### DIFF
--- a/modules/builders/BUILD.bazel
+++ b/modules/builders/BUILD.bazel
@@ -1,0 +1,33 @@
+load("//tools:defaults.bzl", "npm_package", "ts_library")
+
+filegroup(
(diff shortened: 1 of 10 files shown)


modules/bazel-tsconfig-build.json
```diff
@@ -15,7 +15,7 @@
     "sourceMap": true,
     "inlineSources": true,
     "target": "es2015",
-    "lib": ["es2015", "dom"],
+    "lib": ["es2018", "dom"],
```
vikerman: Why was this change needed?

alan-agius4: Not needed now, removing

PR #1265, opened 2019-09-30
refactor: remove output path logic as it's handled in universal schem…

alan-agius4:
refactor: remove output path logic as it's handled in universal schematic

If the outputPath change is needed, this should be done at upstream in `angular/cli` instead

https://github.com/angular/angular-cli/blob/102cd86a8ea33bf1c3dc8cb75b8727c2f239c3d5/packages/schematics/angular/universal/index.ts#L42

### DIFF
--- a/modules/common/schematics/add/index.spec.ts
+++ b/modules/common/schematics/add/index.spec.ts
@@ -41,6 +41,14 @@ describe('Add Schematic Rule', () => {
     expect(productionConfig.optimization).toBeDefined();
   });
 
+  it('should add scripts to package.json', async () => {
+    const tree = await schematicRunner
+      .callRule(addUniversalCommonRule(defaultOptions), appTree).toPromise();
+    const {scripts} = JSON.parse(tree.read('package.json')!.toString());
+    expect(scripts['build:ssr']).toBe('ng build --prod && ng run test-app:server:production');
+    expect(scripts['serve:ssr']).toBe('node dist/test-app/server/main.js');
+  });
+
   it(`should update 'tsconfig.server.json' files with main file`, async () => {
     const tree = await schematicRunner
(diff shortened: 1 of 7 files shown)


alan-agius4:
//cc @clydin would you also mind taking a look at this please?

modules/common/schematics/install/index.spec.ts
```diff
@@ -32,15 +32,24 @@ describe('Universal Schematic', () => {
     const contents = JSON.parse(tree.readContent('angular.json'));
     const architect = contents.projects.bar.architect;
     expect(architect.build.configurations.production).toBeDefined();
-    expect(architect.build.options.outputPath).toBe('dist/bar/browser');
-    expect(architect.server.options.outputPath).toBe('dist/bar/server');
+    expect(architect.build.options.outputPath).toBe('dist/bar');
```
alan-agius4: Note: these tests will need to be changed once https://github.com/angular/angular-cli/pull/15717 is landed and released.

alan-agius4: Note: these tests will need to be changed once https://github.com/angular/angular-cli/pull/15717 is landed and released.

alan-agius4: Note: these tests will need to be changed once https://github.com/angular/angular-cli/pull/15717 is landed and released.

vikerman:
We'll wait for the next release of CLI and check  in with adjusted tests so that they don't break out of context.

vikerman:
@alan-agius4 - The next CLI version has been published. Please update the tests. Thanks!

alan-agius4:
@vikerman PR updated

CaerusKaru:
Needs rebase

alan-agius4:
@CaerusKaru, rebased again

PR #1266, opened 2019-09-30
feat: add migrations for version 9

alan-agius4:
Note: the migrations are currently being tested on via unit tests since there is no real E2E infra in this repo.

### DIFF
--- a/modules/common/schematics/migrations/update-9/index.ts
+++ b/modules/common/schematics/migrations/update-9/index.ts
@@ -10,7 +10,6 @@ import {
   Rule,
   SchematicsException,
   chain,
-  Tree,
   externalSchematic,
 } from '@angular-devkit/schematics';
 import {removePackageJsonDependency} from '@schematics/angular/utility/dependencies';
@@ -28,8 +27,7 @@ export function version9UpdateRule(collectionPath: string): Rule {
 
     return chain([
       removePackageScriptsRule(),
-      backupOldFilesRule(),
-      ... await updateProjectsStructureRule(host, collectionPath),
+      updateProjectsStructureRule(collectionPath),
       (tree, context) => {
         const packageChanges = tree.actions.some(a => a.path.endsWith('/package.json'));
         if (context && packageChanges) {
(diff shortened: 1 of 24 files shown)


alan-agius4:
//cc @clydin would you also mind taking a look at this please?

alan-agius4:
WIP: because I'd like to add some more tests.

modules/common/schematics/migrations/update-9/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    // Remove old scripts in 'package.json'
+    const pkgPath = '/package.json';
+    const buffer = tree.read(pkgPath);
+    if (!buffer) {
+      throw new SchematicsException('Could not find package.json');
+    }
+
+    const pkg = JSON.parse(buffer.toString());
+    delete pkg.scripts['compile:server'];
+    delete pkg.scripts['build:client-and-server-bundles'];
```
CaerusKaru: What happens if the user has customized this target?

vikerman: We should probably not touch the scripts - Or backup the old package.json also.

modules/common/schematics/migrations/update-9/index.ts
```diff
@@ -0,0 +1,96 @@
...
+import {getWorkspace} from '@schematics/angular/utility/workspace';
+import {NodePackageInstallTask} from '@angular-devkit/schematics/tasks';
+import {Builders} from '@schematics/angular/utility/workspace-models';
+import {normalize, join} from '@angular-devkit/core';
+import {Schema as InstallSchema} from '../../install/schema';
+
+export function version9UpdateRule(collectionPath: string): Rule {
+  return async host => {
+    // the below dependencies are no longer needed in version 9
+    removePackageJsonDependency(host, 'ts-loader');
```
vikerman: Same here. Not sure whether we want to remove this without backing up the package.json. Simpler thing would be to just leave these in.

modules/common/schematics/migrations/update-9/index.ts
alan-agius4: Good question, well this target is no longer required nor needed to build universal.

We keep either keep it or remove it, though if we keep it there is no guarantee that it will work if users changed it to depend on `compile:server`

modules/common/schematics/migrations/update-9/index.ts
alan-agius4: I am not quite sure about this, personally I think there git for file versioning and backups!

But if you think it’s better though I’ll do it.

modules/common/schematics/migrations/update-9/index.ts
alan-agius4: If rather backup but, as mentioned there is also git which thinking about it a bit more expect for server.ts backup it might be a bit redundant. As users will require to eventually find and delete these backup files.

alan-agius4:
Comments addressed

PR #1271, opened 2019-10-05
fix(common): update server main option when using `ng add/ng update`

alan-agius4:
The server target main option needs to be updated when running the add schematic to point to the server.ts entrypoint

### DIFF
--- a/modules/common/schematics/add/index.ts
+++ b/modules/common/schematics/add/index.ts
@@ -12,13 +12,14 @@ import {
   SchematicsException,
   noop,
 } from '@angular-devkit/schematics';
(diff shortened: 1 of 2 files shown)


modules/common/schematics/add/index.ts
```diff
@@ -102,17 +121,3 @@ function updateServerTsConfigRule(options: AddUniversalOptions): Rule {
     }
   };
 }
-
-export default function (options: UniversalOptions): Rule {
```
alan-agius4: Oops seems like there was a wrong merge from my side :(

PR #1278, opened 2019-10-14
feat: add tslib as peerDependency

alan-agius4:
feat: add tslib as peerDependency

nguniversal, relied on a hidden dependency of `tslib` due to `importHelpers`
https://github.com/angular/universal/blob/5798f198db1fb0bce9dbfb850751429a3724084a/modules/bazel-tsconfig-build.json#L11 which was previously installed by the Angular framework. This, however changed with this PR https://github.com/angular/angular/pull/32167

With this change we now list `tslib` as a required peerDepedency.

Users not using the CLI will need to manually install `tslib` via;
```
yarn add tslib
```
or
```
npm install tslib --save
```

Reference: TOOL-836


alan-agius4:
Also related to https://github.com/angular/angular/pull/32167

PR #1323, opened 2019-11-04
feat(builders): implement ssr-dev-server

manfredsteyer:
The ssr-dev-server builder starts the browser builder and the server builder in watch mode.
To show how it works, one can run the serve:ssr-dev-server npm script in the express-engine-ivy integration test project

### DIFF
--- a/integration/express-engine-ivy/angular.json
+++ b/integration/express-engine-ivy/angular.json
@@ -95,6 +95,19 @@
             }
           }
         },
+        "ssr-serve": {
+          "builder": "@nguniversal/builders:ssr-dev-server",
+          "options": {
+            "browserTarget": "express-engine-ivy:build",
+            "serverTarget": "express-engine-ivy:server"
+          },
+          "configurations": {
+            "production": {
+              "browserTarget": "express-engine-ivy:build:production",
+              "serverTarget": "express-engine-ivy:server:production"
+            }
+          }
+        },        
         "extract-i18n": {
           "builder": "@angular-devkit/build-angular:extract-i18n",
           "options": {
--- a/integration/express-engine-ivy/package.json
+++ b/integration/express-engine-ivy/package.json
@@ -14,6 +14,7 @@
     "test": "npm run build && npm run serve:test",
     "serve:test": "concurrently \"npm run serve:ssr\" \"npm run e2e\" --kill-others --success first",
     "serve:ssr": "node dist/express-engine-ivy/server/main.js",
+    "serve:ssr-dev-server": "ng run express-engine-ivy:ssr-serve",
     "build:ssr": "npm run build:client-and-server-bundles",
     "build": "ng build --prod && ng run express-engine-ivy:server:production"
   },
@@ -40,6 +41,7 @@
     "@types/express": "file:../../node_modules/@types/express",
     "@types/node": "file:../../node_modules/@types/node",
     "@types/jasmine": "file:../../node_modules/@types/jasmine",
+    "@nguniversal/builders": "file:../../dist/modules-dist/builders",
     "codelyzer": "5.2.0",
     "concurrently": "5.0.0",
     "jasmine-core": "3.5.0",
--- a/integration/hapi-engine-ve/package.json
+++ b/integration/hapi-engine-ve/package.json
@@ -14,6 +14,7 @@
     "test": "npm run build && npm run serve:test",
     "serve:test": "concurrently \"npm run serve:ssr\" \"npm run e2e\" --kill-others --success first",
     "serve:ssr": "node dist/hapi-engine-ve/server/main.js",
+    "serve:ssr-dev-server": "ng run hapi-engine-ve:ssr-serve",
     "build:ssr": "npm run build:client-and-server-bundles",
     "build": "ng build --prod && ng run hapi-engine-ve:server:production"
   },
@@ -42,6 +43,7 @@
     "@types/hapi__inert": "^5.2.0",
     "@types/node": "file:../../node_modules/@types/node",
     "@types/jasmine": "file:../../node_modules/@types/jasmine",
+    "@nguniversal/builders": "file:../../dist/modules-dist/builders",
     "codelyzer": "5.2.0",
     "concurrently": "5.0.0",
     "jasmine-core": "3.5.0",
--- a/modules/builders/BUILD.bazel
+++ b/modules/builders/BUILD.bazel
@@ -38,5 +38,6 @@ ts_library(
 npm_package(
     name = "npm_package",
     srcs = [":builders_assets"],
+    tags = ["release"],
     deps = [":builders"],
 )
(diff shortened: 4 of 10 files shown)


modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -6,18 +6,48 @@
...
 import { Schema } from './schema';
+import {
+  switchMap,
+  map
+} from 'rxjs/operators';
 
 export type SSRDevServerBuilderOptions = Schema & json.JsonObject;
 
 export function execute(
   _options: SSRDevServerBuilderOptions,
```
alan-agius4: Nit: options is no longer unused, thus the _ can be removed

integration/express-engine-ivy/package.json
```diff
@@ -14,6 +14,7 @@
     "test": "npm run build && npm run serve:test",
     "serve:test": "concurrently \"npm run serve:ssr\" \"npm run e2e\" --kill-others --success first",
     "serve:ssr": "node dist/express-engine-ivy/server/main.js",
+    "serve:ssr-dev-server": "ng run express-engine-ivy:ssr-dev-server",
```
alan-agius4: In order to test this, you need to replace `serve:ssr` script, also I’d prefer if we update all integration tests to use the new builder.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -6,18 +6,48 @@
...
+    watch: true,
+    serviceWorker: false,
+  }));
+
+  const serverTargetRun = from(_context.scheduleTarget(serverTarget, {
+    watch: true,
+  }));
+
+  return combineLatest(browserTargetRun, serverTargetRun).pipe(
+    switchMap(([b, s]) => zip(b.output, s.output)),
```
alan-agius4: I think you missed the logic to spawn the server.

manfredsteyer: Correct, that will be the next PR.

integration/express-engine-ivy/package.json
manfredsteyer: Will update the other integration tests too.

Could using ``serve:ssr`` affect the existing tests as we are using watch mode here? My fear is that we are ending up in an infinity loop.

alan-agius4: No it shouldn't

modules/builders/BUILD.bazel
```diff
@@ -38,5 +38,6 @@ ts_library(
 npm_package(
     name = "npm_package",
     srcs = [":builders_assets"],
+    tags = ["release"],
```
alan-agius4: Can we revert this so that we don't release this package until this feature is usable?

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -6,18 +6,48 @@
...
 export function execute(
   _options: SSRDevServerBuilderOptions,
   _context: BuilderContext,
 ): Observable<BuilderOutput> {
-  return of({ success: true });
+
+  const browserTarget = targetFromTargetString(_options.browserTarget);
+  const serverTarget = targetFromTargetString(_options.serverTarget);
+
+  const browserTargetRun = from(_context.scheduleTarget(browserTarget, {
```
alan-agius4: I am worried about the first run being in parallel

1) Both builds will fight for resources
2) Progress bars issues in the console

On the last note, how are we handling that?

manfredsteyer: Don't we have this issue in general for each run or just for the first one?

alan-agius4: For the progress bar it's for all runs, but regarding the other point it mainly effects the first run, because during the other runs, there are fewer IO operations because most of the files would already be in memory.

integration/express-engine-ivy/package.json
manfredsteyer: Ok

modules/builders/BUILD.bazel
manfredsteyer: ok

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: I see. I've taken care about 1)
Regarding 2) I would say let's brainstorm this and come up with another PR.

integration/express-engine-ivy/angular.json
```diff
@@ -95,6 +95,13 @@
             }
           }
         },
+        "ssr-dev-server": {
+          "builder": "../../dist/modules-dist/builders:ssr-dev-server",
+          "options": {
+            "browserTarget": "express-engine-ivy:build",
+            "serverTarget": "express-engine-ivy:server"
+          }
```
alan-agius4: Add a production configuration

integration/express-engine-ivy/package.json
```diff
@@ -13,7 +13,7 @@
     "e2e": "ng e2e",
     "test": "npm run build && npm run serve:test",
     "serve:test": "concurrently \"npm run serve:ssr\" \"npm run e2e\" --kill-others --success first",
-    "serve:ssr": "node dist/express-engine-ivy/server/main.js",
+    "serve:ssr": "ng run express-engine-ivy:ssr-dev-server",
```
alan-agius4: test against production conf

modules/builders/BUILD.bazel
alan-agius4: If you are going to test this on CI, you need the release flag but also you need to implement the serving part.

manfredsteyer: I see. As this all comes with further PRs, I would say, let's reset the serve npm script to the original one.

integration/hapi-engine-ve/package.json
```diff
@@ -42,6 +43,7 @@
     "@types/hapi__inert": "^5.2.0",
     "@types/node": "file:../../node_modules/@types/node",
     "@types/jasmine": "file:../../node_modules/@types/jasmine",
+    "@nguniversal/express-engine": "file:../../dist/modules-dist/express-engine",
```
alan-agius4: This dependency is incorrect it should be on `@nguniversal/builders`

integration/express-engine-ivy/angular.json
alan-agius4: ex
```
"ssr-server": {
    "builder": "@nguniversal/builders:ssr-dev-server",
    "options": {
        "browserTarget": "express-engine-ivy:build",
        "serverTarget": "express-engine-ivy:server"
    },
    "configurations": {
        "production": {
            "browserTarget": "express-engine-ivy:build:production",
            "serverTarget": "express-engine-ivy:server:production"
        }
    }
},
```

integration/express-engine-ivy/package.json
manfredsteyer: works, but is quite slow. Hence, the npm script still points to the non-production build.

PR #1330, opened 2019-11-12
feat(builders): browser-sync for ssr-dev-server

manfredsteyer:
After building both, the server bundles and the client bundles,
the ssr-dev-server starts the node server as well as a
browser-sync server. After the bundles changed, the node
server is restarted and the browser-sync server notifies
the browser.


modules/builders/BUILD.bazel
```diff
@@ -31,7 +31,10 @@ ts_library(
     deps = [
         "@npm//@angular-devkit/architect",
         "@npm//@angular-devkit/core",
+        "@npm//@types/browser-sync",
+        "@npm//browser-sync",
         "@npm//rxjs",
+        "@npm//tree-kill",
```
alan-agius4: Can we not use tree-kill please? This is a bit redundant in this case

modules/builders/src/ssr-dev-server/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+            if (err) {
+                obs.error(err);
+                return;
+            }
+
+            obs.next({ stdout, stderr });
+            obs.complete();
+        });
+
+        return () => treeKill(proc.pid, 'SIGTERM');
```
alan-agius4: Instead of tree-kill use

```
if (!process.killed){
 process.kill();
}

File: modules/builders/src/ssr-dev-server/index.ts [hunk 3]
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+      map(_ => ([browserOutput, serverOutput]))
+    );
+}
+
+function startBrowserSync(port: number): browserSync.BrowserSyncInstance {
+
+  return browserSync.init({
+    proxy: `localhost:${port}`,
+    // TODO: Make this port configurable
+    port: 4200,
```
alan-agius4: Can we do this? It’s just a matter of adding a couple of lines

bazel-universal-fork
```diff
@@ -0,0 +1 @@
+/home/manfred/.cache/bazel/_bazel_manfred/3c5b42f541f2eebfc8b037dd0cc1c8d0/execroot/nguniversal
```
alan-agius4: Revert this file pls

package.json
```diff
@@ -57,12 +57,14 @@
...
     "@hapi/hapi": "^18.4.0",
     "@schematics/angular": "^9.0.0-next.14",
+    "@types/browser-sync": "^2.26.1",
     "@types/express": "^4.17.1",
     "@types/fs-extra": "^8.0.0",
     "@types/hapi__hapi": "^18.2.5",
     "@types/hapi__inert": "^5.2.0",
     "@types/jasmine": "^3.4.4",
     "@types/node": "^12.11.1",
+    "browser-sync": "^2.26.7",
```
alan-agius4: This should be added as a dependency of the builders package.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+      });
+    }),
   );
 }
 
+function startNodeServer(browserOutput: BuilderOutput, serverOutput: BuilderOutput, port: number) {
+  const path = serverOutput.outputPath;
+  const env = { ...process.env, PORT: '' + port };
+
+  return execAsObservable(`node ${path}/main.js`, { env })
```
alan-agius4: The path here needs to be escaped due to special characters.

If I recall correctly this path is absolute. Can you confirm?

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+    catchError(error => {
+      return of({
+        success: false,
+        error: error
+      });
+    }),
   );
 }
 
+function startNodeServer(browserOutput: BuilderOutput, serverOutput: BuilderOutput, port: number) {
```
alan-agius4: BrowserOutput is redundant here.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+  const path = serverOutput.outputPath;
+  const env = { ...process.env, PORT: '' + port };
+
+  return execAsObservable(`node ${path}/main.js`, { env })
+    .pipe(
+      startWith({ stdout: '', stderr: '' }),
+        // Emit a signal after the process has been started
+      map(_ => ([browserOutput, serverOutput]))
```
alan-agius4: Why not map in the main function.

In here we also need to handle server errors, if the server emits and error we shouldn’t crash.

Also probably stdout should be pipped to avoid having log messages in each restart such as `node server is listening on port 3333`

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+    );
+}
+
+function startBrowserSync(port: number): browserSync.BrowserSyncInstance {
+
```
alan-agius4: We need to do some logs like the port the server is listing too, etc. Similar to `ng serve`.

Will get you the message later.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+  return from(getAvailablePort()).pipe(
+    tap(port => {
+      nodeServerPort = port;
+    }),
+    switchMap(_ => merge(browserTargetRun, serverTargetRun, 1)),
+    bufferCount(2),
     switchMap(([b, s]) => zip(b.output, s.output)),
-    map(([b, s]) => ({ success: b.success && s.success }))
+    switchMap(([b, s]) => startNodeServer(b, s, nodeServerPort)),
```
alan-agius4: If both builds weren’t successful we should try to spawn the server

modules/builders/src/ssr-dev-server/index.ts
alan-agius4:
```js
            context.logger.info(tags.oneLine`
              **
              Angular Universal Live Development Server is listening on http://localhost:${port},
              open your browser on http://localhost:${port}
              **
            `);

```

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
-    pairwise(),
+  let bsInstance: browserSync.BrowserSyncInstance;
+  let nodeServerPort: number;
+
+  return from(getAvailablePort()).pipe(
+    tap(port => {
+      nodeServerPort = port;
+    }),
+    switchMap(_ => merge(browserTargetRun, serverTargetRun, 1)),
+    bufferCount(2),
```
alan-agius4: This is not fully correct, a user can do a change which triggers only a server builder or client build, example changing the `server.ts`.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+    switchMap(([b, s]) => startNodeServer(b, s, nodeServerPort)),
+    tap(_ => {
+      if (!bsInstance) {
+        bsInstance = startBrowserSync(nodeServerPort);
+      } else {
+        bsInstance.reload();
+      }
+    }),
+    map(([b, s]) => ({ success: b.success && s.success })),
+    catchError(error => {
```
alan-agius4: Nit 
```js
    catchError(error => {
      return of({
        success: false,
        error: error
      });
    }),
```

```js
    catchError(error => of({
        success: false,
        error,
     });
 ),

```

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
     watch: true,
   }));
 
-  return merge(browserTargetRun, serverTargetRun, 1).pipe(
-    pairwise(),
+  let bsInstance: browserSync.BrowserSyncInstance;
+  let nodeServerPort: number;
+
+  return from(getAvailablePort()).pipe(
+    tap(port => {
```
alan-agius4: Nit: 

```js
  tap(port => nodeServerPort = port),
```

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: +1

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: good point

Yes, it's absolute.

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: What do you mean? (browserOuput vs serverOutput)

modules/builders/BUILD.bazel
manfredsteyer: You mean `tree-kill` instead of `@npm//tree-kill` ?

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: IMHO it's ok in this case, b/c this very switchMap only runs once at the start. After each change, the other ones run.

Do I see this right?

However, how to deal with this special case within the other switchMaps? Using `zip` could be an issue in cases where only one build changes. Right?

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: What do you mean?

modules/builders/BUILD.bazel
alan-agius4: No, remove tree-kill as a dependency all together

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: the browserOutput object here is redundant., it is used by the `map()` but really we don't need that to be done here.

alan-agius4: See comment below.

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: 1) If the server crashes, we should crash the watch, so we need to gracefully catch the error
2) if the server throw an error, we need to handle the stderr and print it in the console
3) I don't think that we should do this in these method: `map(_ => ([browserOutput, serverOutput]))` strictly speaking this function doesn't care about the browserOutput and providing such imports for mapping seems like an overkill and not the best in terms of maintainability to some extent. I think a better and cleaner approach is to use `pipe`.

```ts
function startNodeServer(outputPath: string, port: number): Observable<void> {
```


Some preudo code;

```
from(getAvailablePort()).pipe(
    tap(port => nodeServerPort = port),
    switchMap(_ => merge(browserTargetRun, serverTargetRun, 1)),
    bufferCount(2),
    switchMap(([b, s]) => zip(b.output, s.output)),
    map(([b, s]) => ({
        success: b.success && s.success,
        error: b.error || s.error,
        serverOutputPath: b.outputPath || '',
    }),
    switchMap(buildEvent => {
        if (!buildEvent.success) {
            return buildEvent;
        }

        return startNodeServer(buildEvent.serverOutputPath, nodeServerPort)
            .pipe(
                tap(() => {
                    if (!bsInstance) {
                        bsInstance = startBrowserSync(nodeServerPort);
                    } else {
                        bsInstance.reload();
                    }
                }),
                // If a server error occuires we should print it but we shouldn't crash the watch
                catchError(() => NEVER),
                mapTo(buildEvent)
            );
    }),
    catchError(error => {
        return of({
            success: false,
            error: error
        });
    }),
);

```

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -45,11 +52,59 @@ export function execute(
...
+      nodeServerPort = port;
+    }),
+    switchMap(_ => merge(browserTargetRun, serverTargetRun, 1)),
```
alan-agius4: Question: why `_` instead of `()` is this a personal preference?

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: You are absolutely right. I think to overcome this we need to do something a bit different in terms of the RXJS patterns used here which would also simplify the logic.

My idea would be to do something along;
```ts
 // the below async parts might make sense to be moved to an init function or something along that route
    const browserTarget = targetFromTargetString(options.browserTarget);
    const serverTarget = targetFromTargetString(options.serverTarget);
  
    const browserTargetRun = await context.scheduleTarget(browserTarget, {
      watch: true,
      serviceWorker: false,
    }));
  
    const serverTargetRun = await context.scheduleTarget(serverTarget, {
      watch: true,
    });
  
    const nodeServerPort = await getAvailablePort();

    const bsInstance = browserSync.init({
        // let's no force refreshes close to each other
        reloadDebounce: 250
    });

    const serverRunOutput =  serverRun.output.pipe(
        // check if success
        // if not don't restart
        tap(() => startNodeServer(b, s, nodeServerPort),
    );

    return combineLatest(
        serverRunOutput,
        browserRun.output
    )
    .pipe(
        // if both success otherwise don't
        tap(() => bsInstance.reload());
    );
```


NB;
- With the above we should also stop the jobs
```ts
await Promise.all([browserTargetRun.stop(), serverTargetRun.stop()]);
```


.

modules/builders/BUILD.bazel
manfredsteyer: If I remove it, I'll get this:

```
modules/builders/src/ssr-dev-server/utils.ts:11:27 - error TS2307: Cannot find module 'tree-kill'.

11 import * as treeKill from 'tree-kill';
                             ~~~~~~~~~~~
```

alan-agius4: Yeah you need to refactor the code to not use tree-kill see, comments above.

alan-agius4: https://github.com/angular/universal/pull/1330#discussion_r345536651

manfredsteyer: Ah, now I see. Thx.

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: No specific reason. I personally prefere ``_``. Shall I change it to ``()``?

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: I've implemented this idea in a minimally invasive way: ``combineLatest`` instead of ``zip`` + ``debounceTime``.

This is also beneficial for the case where the user is saving several files in a row.

My experiments also show that starting/ reloading browserSync when the node server is not fully available leads to issues (at least under windows). This is also solved by a short ``debounceTime``.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+        // Emit a signal after the process has been started
+    );
+}
+
+function startBrowserSync(
+  syncServerPort: number,
+  nodeServerPort: number,
+  logger: logging.LoggerApi): Observable<browserSync.BrowserSyncInstance> {
+
+  const syncServerPort$ = (syncServerPort === 0) ?
```
alan-agius4: I would strongly recommend in not using rxjs for this stuff. It’s an overkill as really these is no reason why this method needs to be reactive.

Async/await will suffice.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+    tap(([b, s]) => {
+      serverTargetRun = s;
+      browserTargetRun = b;
+    }),
+    switchMap(([b, s]) => combineLatest(b.output, s.output)),
+    switchMap(([b, s]) => {
+      if (!b.success || !s.success) {
+        return of([b, s]);
+      }
+      return startNodeServer(s, nodeServerPort).pipe(
```
alan-agius4: Would this start the node server twice for every change? Ie: when the browser and server is changed.


Hence, that was the reason why in my example is left the streams separate.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+            return startBrowserSync(options.port, nodeServerPort, context.logger).pipe(
+              tap(instance => bsInstance = instance),
+              mapTo([b, s])
+            );
+          } else {
+            bsInstance.reload();
+            return of([b, s]);
+          }
+        }),
+        catchError(err => {
```
alan-agius4: The catch error should be for the nodeServer part since If that throws we don’t want to continue with browser sync etc...

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+          console.error('Error in browser sync', err);
```
alan-agius4: Don’t use console logs, use the logger from context.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+      if (!b.success || !s.success) {
+        return of([b, s]);
+      }
+      return startNodeServer(s, nodeServerPort).pipe(
+        tap(result => {
+          if (result.stderr) {
+            console.error('Error in node server', result.stderr);
+          }
+        }),
+        debounceTime(500),
```
alan-agius4: This is rather a high debounce time, why is this needed?

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+            return of([b, s]);
+          }
+        }),
+        catchError(err => {
+          console.error('Error in browser sync', err);
+          return NEVER;
+        }),
+      );
+    }),
+    map(([b, s]) => ({ success: b.success && s.success })),
```
alan-agius4: You are not handling errors in this map, see previous example: 
See: https://github.com/angular/universal/pull/1330#discussion_r346191990

```ts
    map(([b, s]) => ({
        success: b.success && s.success,
        error: b.error || s.error,
        serverOutputPath: b.outputPath || '',
    }),
```

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +46,126 @@ export function execute(
...
+    }),
+    switchMap(([b, s]) => combineLatest(b.output, s.output)),
+    switchMap(([b, s]) => {
+      if (!b.success || !s.success) {
+        return of([b, s]);
+      }
+      return startNodeServer(s, nodeServerPort).pipe(
+        tap(result => {
+          if (result.stderr) {
+            console.error('Error in node server', result.stderr);
```
alan-agius4: Having an `stderr` not empty doesn't necessary mean there was error.
stderr, can be used for various reasons;

1) error
2) warnings
3) progress indicators

etc..

alan-agius4: don't use console.logs, use the logger in context

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: If you don't mind :) Since that's typically the common pattern used in our repos.

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: Sorry I just realised I did it wrong in here https://github.com/angular/universal/pull/1330#discussion_r346191990 but correct in here https://github.com/angular/universal/pull/1330#discussion_r346245761 since it's outside of rxjs

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: +1 Will do it in a new PR.

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: Now, we only restart the node server if the server bundle is rebuilt.

bazel-universal-fork
manfredsteyer: it's reverted now.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
...
+    })),
+    finalize(() => {
+      serverTargetRun.stop();
+      browserTargetRun.stop();
+    })
   );
 }
 
+function startNodeServer(serverOutput: BuilderOutput, port: number) {
+  const path = normalize(serverOutput.outputPath + '/main.js');
```
alan-agius4: `join` is better in this case.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
...
+    switchMap(([br, sr]) => {
+      const server$ = sr.output.pipe(
+        switchMap(s => {
+          if (s.error) {
+            return of(s);
+          }
+          return startNodeServer(s, nodeServerPort).pipe(
+            map(() => s),
+            catchError(err => {
+              context.logger.error('Error in browser sync', err);
```
alan-agius4: The error message here doesn’t seem right as this is unrelated to browser sync.

Maybe better something like 
`A server error has occurred.`

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
...
+    tap(port => nodeServerPort = port),
+    switchMap(() => combineLatest(browserTargetRun$, serverTargetRun$)),
+    tap(([br, sr]) => {
+      serverTargetRun = sr;
+      browserTargetRun = br;
+    }),
+    switchMap(([br, sr]) => {
+      const server$ = sr.output.pipe(
+        switchMap(s => {
+          if (s.error) {
```
alan-agius4: You shouldn’t check `.error` check `.success` instead. We might have an unsuccessful build  without an error msg.

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
...
+        return startBrowserSync(options.port, nodeServerPort, context.logger).pipe(
+          tap(instance => bsInstance = instance),
+          mapTo([b, s])
+        );
+      } else {
+        bsInstance.reload();
+        return of([b, s]);
+      }
+    }),
+    map(([b, s]) => ({
```
alan-agius4: This should be at the very being not at the very end. So that you avoid adding doing code like 

`if (!b.success || !s.success)`

Or 

`of([b, s])`

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
...
-    pairwise(),
-    switchMap(([b, s]) => zip(b.output, s.output)),
-    map(([b, s]) => ({ success: b.success && s.success }))
+  let bsInstance: browserSync.BrowserSyncInstance;
+  let nodeServerPort: number;
+  let serverTargetRun: BuilderRun;
+  let browserTargetRun: BuilderRun;
+
+  return from(getAvailablePort()).pipe(
+    tap(port => nodeServerPort = port),
```
alan-agius4: Move this after the build, the more time you want to assign the port the greater the risk it will become unavailable.

bazel-universal-fork
alan-agius4: Bad file!!! Keeps re-appearing!!!

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: Since you are on Windows can you check that it works on paths with spaces and special characters such as `(` without escaping the path?

modules/builders/src/ssr-dev-server/index.ts
manfredsteyer: As it is currently a big stream, I think I need it at the end. Before, I need ``[b, s]`` in the individual steps.

Or do I miss sth here?

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
   const browserTarget = targetFromTargetString(options.browserTarget);
   const serverTarget = targetFromTargetString(options.serverTarget);
 
-  const browserTargetRun = from(context.scheduleTarget(browserTarget, {
+  const browserTargetRun$ = from(context.scheduleTarget(browserTarget, {
     watch: true,
     serviceWorker: false,
   }));
```
alan-agius4: Can you also add `extractCss: true` here please

//cc @vikerman

modules/builders/src/ssr-dev-server/index.ts
alan-agius4: You don't need `[b, s]` in the steps :)

```ts
return from(getAvailablePort()).pipe(
    tap(port => nodeServerPort = port),
    switchMap(() => combineLatest(browserTargetRun$, serverTargetRun$)),
    tap(([br, sr]) => {
        serverTargetRun = sr;
        browserTargetRun = br;
    }),
    switchMap(([br, sr]) => {
        const server$ = sr.output.pipe(
            switchMap(s => {
                if (s.error) {
                    return of(s);
                }
                return startNodeServer(s, nodeServerPort).pipe(
                    map(() => s),
                    catchError(err => {
                        context.logger.error('Error in browser sync', err);
                        return NEVER;
                    }),
                );
            }));

        return combineLatest(br.output, server$);
    }),
    map(([b, s]) => ({
        success: b.success && s.success,
        error: b.error || s.error,
        serverOutputPath: b.outputPath || '',
    } as BuilderOutput)),
    switchMap(buildEvent => {
        if (!bsInstance && buildEvent.success) {
            return startBrowserSync(options.port, nodeServerPort, context.logger).pipe(
                tap(instance => bsInstance = instance),
                mapTo(buildEvent),
            );
        }

        if (buildEvent.success) {
            bsInstance.reload();
        }
        return of(buildEvent);
    }),
    catchError(error => of({
        success: false,
        error: mapErrorToMessage(error)
    })),
    finalize(() => {
        serverTargetRun.stop();
        browserTargetRun.stop();
    })
);
```

modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -36,20 +45,132 @@ export function execute(
...
+    map(([b, s]) => ({
+      success: b.success && s.success,
+      error: b.error || s.error,
+      serverOutputPath: b.outputPath || '',
+    } as BuilderOutput)),
+    catchError(error => of({
+        success: false,
+        error: mapErrorToMessage(error)
+    })),
+    finalize(() => {
```
alan-agius4: Since you are using obs, I think you don't need to stop since it should be handled in the tear down logic. I'll check tmr.

modules/builders/src/ssr-dev-server/schema.json
```diff
@@ -11,6 +11,11 @@
     "serverTarget": {
       "type": "string",
       "description": "Browser target to build."
+    },
+    "port": {
+      "type": "number",
+      "default": 4200,
+      "description": "Port to start the sync server at. Default is 4200. Pass 0 to get a dynamically assigned port"
```
alan-agius4: Nit: missing full stop at the end

bazel-universal-fork
manfredsteyer: Should be gone now. If you click at the file name above, the link is broken.

alan-agius4: Still here and the link is not broken

alan-agius4:
@googlebot I consent.

integration/express-engine-ivy/package.json
```diff
@@ -10,11 +10,11 @@
     "postinstall": "webdriver-manager update --gecko false --standalone false $CHROMEDRIVER_VERSION_ARG",
     "ng": "ng",
     "start": "ng serve",
-    "e2e": "ng e2e",
-    "test": "npm run build && npm run serve:test",
-    "serve:test": "concurrently \"npm run serve:ssr\" \"npm run e2e\" --kill-others --success first",
+    "e2e": "sleep 60 && ng e2e",
```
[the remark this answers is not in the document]
alan-agius4: The sleep is needed because since now we build and run e2e concurrently e2e will run prior to the server being started.

While I did a method to probe for the app-root, I did the sleep to wait prior to even start this. https://github.com/angular/universal/pull/1330/files#diff-73654ffb2ae5aa618d303608b80b9775R15

integration/express-engine-ivy/package.json
```diff
@@ -10,11 +10,11 @@
...
+    "test": "npm run serve:test",
+    "serve:test": "concurrently \"npm run serve:ssr:dev\" \"npm run e2e\" --kill-others --success first",
     "serve:ssr": "node dist/express-engine-ivy/server/main.js",
-    "serve:ssr-dev-server": "ng run express-engine-ivy:ssr-serve",
+    "serve:ssr:dev": "ng run express-engine-ivy:serve-ssr:production",
```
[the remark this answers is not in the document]
alan-agius4: That's just for the integration tests since we run e2e integration tests in production. we can rename the `script` name. 

Maybe for test purposes we can name it 
```
serve:ssr:prod
```

PR #1350, opened 2019-12-03
chore: update deps

vikerman:


alan-agius4:
Can you also update this list https://github.com/angular/universal/blob/8c2aa0580ba62c7e61892ffeceda4bb4a74791a9/tools/defaults.bzl#L35?

PR #1353, opened 2019-12-03
fix(builders): remove cache-control header from responses

alan-agius4:


vikerman:
I've seen sometimes images flicker when caching is disabled for them. The initial ssr displays the image - but image is redownloaded when client bootstraps - causing a flicker.

Should we disable cache only for certain resources?(js, html, css)

alan-agius4:
As per our offline convo, lets remove the cache control header and rely on etags. This is the same behaviour of `ng serve`.

That is of course if this doesn't cause flickering, which in my test case seems it doesn't.

PR #1360, opened 2019-12-05
fix(builder): add option to provide a custom host

alan-agius4:


modules/builders/src/ssr-dev-server/schema.ts
```diff
@@ -13,15 +13,18 @@ export interface Schema {
...
 
+  /** Host to listen on. */
+  host?: string;
+
   /**
    * Port to start the sync server at.
    * Default is 4200. Pass 0 to get a dynamically assigned port.
    */
-  port: number;
+  port?: number;
```
vikerman: Just a clarification. If there is a default value should this still have the ?

alan-agius4: Yeah, cause the property itself is optional when using the api. The default will be applied the the schema transformer

vikerman: But that means the code using this will surely have a port right? So it would be redundant to check whether port is defined before using it in the TS code.

vikerman: Alan checked with quicktype and that adds the ?. So that's the right thing to do here.

PR #1370, opened 2019-12-09
chore: update deps

vikerman:


CaerusKaru:
@vikerman Just needs updated lockfile

PR #1382, opened 2019-12-11
fix(common): add production config section to prerender builder config

vikerman:
This can help the firebase deploy builder(for instance) to call the either the browser build or prerender target uniformly with configuration set to production.

### DIFF
--- a/modules/common/schematics/add/index.ts
+++ b/modules/common/schematics/add/index.ts
@@ -118,7 +118,12 @@ function updateWorkspaceConfigRule(options: AddUniversalOptions): Rule {
           browserTarget: `${projectName}:build:production`,
           serverTarget: `${projectName}:server:production`,
           routes: []
-        }
+        },
+        // Add a dummy production config to be consistent with other targets.
+        configurations: {
+          production: {
+          },
+        },
       });
     });
   };


alan-agius4:
firebase builder only works with `build` target, so this won’t make it possible to use it with firebase.

Also typically ng deploy builders also run a single builder which is typically the browser builders and deploy the output of that. Running another builder and deploy the result of that will get you a broken journey because for example you will deploy the prerendered application but you won’t have the assets as they will be emitted by the browser builder 

https://github.com/angular/angularfire/blob/76f721254bcb4e9db693486332a4a72a7e897f15/src/schematics/deploy/actions.ts#L29

vikerman:
@alan-agius4 - The idea is that we can make the target configurable for the deploy builders instead of always calling "build" similar to the azure one - https://github.com/Azure/ng-deploy-azure/blob/8c391d3d951c59c3d2a5e8b83566462a1af39321/src/builders/actions/deploy.ts#L68

If the target can be changed to "prerender" from "build" using the deploy config, ng deploy will build the static site and deploy it,

alan-agius4:
In that case wouldn’t it make sense to provide a  “configuration” value as well?

vikerman:
The production config is same as default config which already points to production versions of browser and server builds. It's being added just for uniformity and convenience

alan-agius4:
Okay, let’s add it, but I still think it should be the builder the handles which configuration target to deploy.

PR #1408, opened 2019-12-19
build: use bazel `disk_cache`

alan-agius4:


alan-agius4:
//cc @gregmagolan

alan-agius4:
Screenshot 2019-12-19 at 23 26 54

PR #1418, opened 2019-12-23
ci: improve circle ci setup

alan-agius4:
With this change we introduce a couple of changes

- Do `yarn install` and code checkout once in the setup phase
- Save cache in `build-packages-dist` to ensure that we cache Bazel action outputs at the very end of our builds so that we will later get better cache hit rate
- We use circle ci 2.1 features such as action executors and commands
- Removed the `bazel-lint` task which was redundant
- We save yarn cache to circli ci cache instead of node_modules directory
- Use a cache key fallback, so to use the last cached version when a dependency is updated.


alan-agius4:
@vikerman can you please remove `bazel-lint` from being a required task in github?

PR #1450, opened 2020-01-15
schematics compatible with IIS Node and Azure App service

alan-agius4:
The problem is that the `interceptor.js` which is used by IIS Node will require the application files rather than executing it. https://github.com/tjanczuk/iisnode/blob/8657944f8803d84514c0c7ddd48c046aa68e6edf/src/scripts/interceptor.js#L210

Thus, the server `run` method will never get invoked.


alan-agius4:
@CaerusKaru, Yeah, though that is not as trivial to do, since we need to setup a windows ci with iis node.

Issue #1459, opened 2020-01-21
Express dev:ssr - Host header is not pass to server


havran:
# 🐞 Bug report

#### What modules are related to this issue?

- [x] express-engine

### Description

I'm trying implement simple Angular service, which should return hostname. If i start `dev:ssr` i always get on server side hostname `localhost`, even if i set header `X-Forwarded-Host`.  

If i build application normally, with `build:ssr`, all works as expected.

Probably server, which proxying  requests to express server, should pass `Host` header through `X-Forwarded-Host` header?

## 🔬 Minimal Reproduction

Service should return correct hostname on server side.

```typescript
import { Inject, Injectable, Optional, PLATFORM_ID } from '@angular/core';
import { isPlatformServer } from '@angular/common';
import { Request } from 'express';
import { REQUEST } from '@nguniversal/express-engine/tokens';

@Injectable({
  providedIn: 'root'
})
export class HostnameService {

  constructor(
    @Inject(PLATFORM_ID) private platformId: Object,
    @Optional() @Inject(REQUEST) protected request: Request
  ) {}

  public getHostname(): string {
    return isPlatformServer(this.platformId) ? this.request.hostname : window.location.hostname;
  }

}
```

## 🌍 Your Environment
<pre><code>

Angular CLI: 9.0.0-rc.9
Node: 10.16.3
OS: linux x64

Angular: 9.0.0-rc.9
... animations, cli, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... platform-server, router
Ivy Workspace: Yes

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.900.0-rc.9
@angular-devkit/build-angular      0.900.0-rc.9
@angular-devkit/build-ng-packagr   0.900.0-rc.9
@angular-devkit/build-optimizer    0.900.0-rc.9
@angular-devkit/build-webpack      0.900.0-rc.9
@angular-devkit/core               9.0.0-rc.9
@angular-devkit/schematics         9.0.0-rc.9
@ngtools/webpack                   9.0.0-rc.9
@nguniversal/builders              9.0.0-rc.0
@nguniversal/common                9.0.0-rc.0
@nguniversal/express-engine        9.0.0-rc.0
@schematics/angular                9.0.0-rc.9
@schematics/update                 0.900.0-rc.9
ng-packagr                         9.0.0-rc.7
rxjs                               6.5.4
typescript                         3.6.4
webpack                            4.41.2
[...]

PR #1462, opened 2020-01-22
fix(builders): send `X-Forwarded` headers when using ssr-dev-server

alan-agius4:
Fixes: #1459


alan-agius4:
CI is red due to https://github.com/angular/universal/pull/1458

PR #1464, opened 2020-01-23
test: disable flaky ci test

alan-agius4:
While this works locally, on ci this is causing 90% of our builds to be red,.


alan-agius4:
I am going to block this for now, to check if it's still a problem with the updated Bazel and NodeJS rules. As it seems to have passed on the run the first run.

alan-agius4:
This test is flakey once again with the new rules https://github.com/angular/universal/pull/1473

PR #1498, opened 2020-02-10
Change ServerAppModule to AppServerModule

wmeints:


wmeints:
@google-admin Signed CLA!

PR #1644, opened 2020-04-29
@nguniversal/builders package improvements

williamdes:
- build: add repository and homepage fields on package.json

I assumed you merge 9.1.x into master afterwards, right ?

### DIFF
--- a/modules/builders/package.json
+++ b/modules/builders/package.json
@@ -2,6 +2,7 @@
   "name": "@nguniversal/builders",
   "version": "0.0.0-PLACEHOLDER",
   "description": "Angular Universal builders package",
+  "homepage": "https://github.com/angular/universal#readme",
   "license": "MIT",
   "keywords": [
     "ssr",
@@ -10,6 +11,13 @@
     "builder",
     "cli"
   ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/angular/universal"
+  },
+  "bugs": {
+    "url": "https://github.com/angular/universal/issues"
+  },
   "main": "src/index.js",
   "typings": "src/index.d.ts",
   "builders": "builders.json",


williamdes:
@googlebot I signed it!

williamdes:
@alan-agius4 I could not fix the integration tests.
Anyway I did not break them

alan-agius4:
@williamdes, the integration test are broken due to the Node.Js update. That said, I don’t see any reason why this should be updated.

modules/builders/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@nguniversal/builders",
   "version": "0.0.0-PLACEHOLDER",
   "description": "Angular Universal builders package",
+  "homepage": "https://github.com/williamdes/universal/tree/master/modules/builders#readme",
```
alan-agius4, line 5: This is not the correct url.

modules/builders/package.json
```diff
@@ -10,6 +11,10 @@
     "builder",
     "cli"
   ],
+  "repository": {
```
alan-agius4, line 14: Can you add the following:

```js
"repository": {
    "type": "git",
    "url": "https://github.com/angular/universal"
  },
  "bugs": {
    "url": "https://github.com/angular/universal/issues"
  },
  "homepage": "https://github.com/angular/universal"
```

alan-agius4:
@williamdes, can you actually create a separate PR for master?

Thanks.

modules/builders/package.json
williamdes: Oh snap !

modules/builders/package.json
williamdes: Done, and I fixed my copy paste URL mistake

williamdes:
Done as #1647

alan-agius4:
> @williamdes, the integration test are broken due to the Node.Js update. That said, I don’t see any reason why this should be updated.

This is still pending, can you drop the commit for CI update in both PRs?

williamdes:
> > @williamdes, the integration test are broken due to the Node.Js update. That said, I don’t see any reason why this should be updated.
> 
> This is still pending, can you drop the commit for CI update in both PRs?

I can but it is needed, as you can see on my commit message
I will do it in ~9h after sleeping

alan-agius4:
I am not seeing any move-file in our CI, I might have missed it though, Can you please elaborate in which step and phase is this happening or share the CIrcle CI build with the exception?

williamdes:
> I am not seeing any move-file in our CI, I might have missed it though, Can you please elaborate in which step and phase is this happening or share the CIrcle CI build with the exception?

The CI failed to install normally because the package move-file needs node +1 minor from ours

See: https://github.com/sindresorhus/move-file/blob/63bf2f051bb6a885d567928d139be6b2a9e9bfaf/package.json#L14

So I updated the version but you say it broke the integration tests, up to you to let me know what to do

williamdes:
I could not find the failing install on mobile, I will remove the commit and see what happens
Maybe it was a side effect

williamdes:
I was right
```
error move-file@2.0.0: The engine "node" is incompatible with this module. Expected version ">=10.17". Got "10.16.3"
error Found incompatible module.
```
https://circleci.com/gh/angular/universal/6730?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

alan-agius4:
Most likely we need to update the lock file, since this issue is not present on master which is using NodeJs 10.16.

williamdes:
Can you help me know what to update, or I will investigate in 4 hours :)

williamdes:
I will have no more time for this now, feel free to cherry-pick my commit from master branch

PR #1656, opened 2020-05-03
chore: Improve prerender logging

dylanwatsonsoftware:
When debugging prerender issues, its often difficult to tell which routes have actually been found.
It seem prudent just to log the routes here.

### DIFF
--- a/modules/builders/src/prerender/utils.ts
+++ b/modules/builders/src/prerender/utils.ts
@@ -45,10 +45,12 @@ export async function getRoutes(
             .filter(route => !route.includes('*') && !route.includes(':'))
         );
       } catch (e) {
-        context.logger.error('Unable to extract routes from application.', e);
+        context.logger.error('Unable to guess the routes from application.', e);
       }
     }
   }
+  
+  context.logger.info(`Found routes: ${routes.join(', ')}`);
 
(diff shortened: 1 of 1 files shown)


dylanwatsonsoftware:
@googlebot I signed it!

CaerusKaru:
I like this change a lot! Can you rebase so we can get this in for the next release?

CaerusKaru:
Also, please change commit message to:

`fix(builders): improve prerender logging`

modules/builders/src/prerender/utils.ts
```diff
@@ -45,11 +45,13 @@ export async function getRoutes(
             .filter(route => !route.includes('*') && !route.includes(':'))
         );
       } catch (e) {
-        context.logger.error('Unable to extract routes from application.', e);
+        context.logger.error('Unable to guess the routes from application.', e);
       }
     }
   }
 
+  context.logger.info(`Found routes: ${routes.join(', ')}`);
```
alan-agius4, line 53: If you have hundreds of routes these logs here might be too verbose.

I reckon, this should be only printed when using the a verbose flag.

alan-agius4:
Closing as PR is staled

PR #1696, opened 2020-05-19
Update tslib and typescript to 3.9

alan-agius4:


alan-agius4:
Blocked on https://github.com/angular/angular-cli/pull/17760

modules/aspnetcore-engine/package.json
```diff
@@ -16,14 +16,13 @@
     "ssr",
     "universal"
   ],
-  "engines": {
-    "node": ">=10.13.0"
```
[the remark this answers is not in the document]
alan-agius4: Mistake 😱

modules/bazel-tsconfig-test.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./bazel-tsconfig-build.json",
   "compilerOptions": {
+    "target": "es5",
```
kyliau, line 4: Could you please explain why this change?
Was `es5` already the default before?
or put it another way, could we use es2015?

kyliau:
Let's cut rc.0 on Tuesday May 26 with this change :)

modules/bazel-tsconfig-test.json
alan-agius4: I actually thought I didn’t commit this.

alan-agius4: Good catch @kyliau.

Issue #1023, opened 2018-06-12
Can universal provide koa-engine?


3fuyu:
Express-engine is good. But we use koa to build project. Can you provide koa-engine? Like #777 .

Toxicable:
We're happy to take a PR for one if you like.
Either a new one or continue on from that one you linked should suffice.

One point of advise would be to use the Common Engine found here: https://github.com/angular/universal/tree/master/modules/common/engine
And here is an example of implementing it: https://github.com/angular/universal/pull/999/files#diff-1d1f74cba58d729d80bea48694d3cd53R33

qqaimh:
Does the koa-engine run perfectly?

JomaKar:
I've seen the PR and I just wanna say there's a beautiful community happy to get news about this topic!

Roman-Lo:
Awesome!!!!!

alan-agius4:
Kindly see https://github.com/angular/universal/pull/1714#issuecomment-745229963

PR #1714, opened 2020-06-05
feat(koa-engine): add koa engine

kgajera:
Closes: #1023

This adds an engine for Koa. Let me know if any changes are needed.


CaerusKaru:
Personally, I like this change, and thanks for contributing! What it comes down to is whether or not the Universal team wants to take ownership of something like this. We'll discuss this internally and reply on this thread. An alternative you may want to consider if we can't pick this up is publishing this package on your own, since it looks like you have all the pieces in place.

alan-agius4:
Thanks for this PR and sorry for the late reply.

During our last team meeting we have discussed that instead of adding additional engines we should make the common engine API public. This will allow the community to create and publish Universal engine wrappers such as this one.

Thanks again.

PR #1780, opened 2020-07-31
Update transfer_http.ts

ansergei:
caching GET request with query string with array of parameter gets the first value only.
it breaks caching for the same url with different query string for example
for api http://api.example.com?params=1&params=2&params=3 cache key will be http://api.example.com?params=1
for api http://api.example.com?params=1&params=2 cache key will be the same http://api.example.com?params=1
and therefor API request will not be sent to server.


ansergei:
@googlebot I signed it!

PR #1816, opened 2020-08-25
build: update to TypeScript 4.0

alan-agius4:


alan-agius4:
Blocked on a CLI release.

PR #1860, opened 2020-10-10
feat: enable absolute URL handling by default for all engines

CaerusKaru:
This functionality was added in Angular v10.1 as an opt-in for
all users. Since ngExpressEngine and ngHapiEngine are meant to
enable best practices out of the box, we enable this for users
by default as well.

### DIFF
--- a/modules/common/engine/src/engine.ts
+++ b/modules/common/engine/src/engine.ts
@@ -49,7 +49,8 @@ export class CommonEngine {
         provide: INITIAL_CONFIG,
         useValue: {
           document: doc,
-          url: opts.url
+          url: opts.url,
+          useAbsoluteUrl: true,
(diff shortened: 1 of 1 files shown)


modules/common/engine/src/engine.ts
```diff
@@ -49,7 +49,8 @@ export class CommonEngine {
         provide: INITIAL_CONFIG,
         useValue: {
           document: doc,
-          url: opts.url
+          url: opts.url,
```
alan-agius4, line 52: I think we might want to provide the users to turn the option off.

If I recall correctly, that was one of the reason this option exists in platform-server in the first place.

alan-agius4:
@CaerusKaru, should we close this PR?

PR #1864, opened 2020-10-17
Updated the express engine to use the absolute url flag

Dig-Doug:
Flag is defined [here](https://github.com/CaerusKaru/angular/blob/98c047b76360218e0b910298416eb1c56b56f163/packages/platform-server/src/tokens.ts#L35). [Background](https://github.com/angular/angular/pull/37539).

The docs [here](https://angular.io/guide/universal#using-absolute-urls-for-http-data-requests-on-the-server) mention that it's handled automatically, but it does seem to work without this flag.

### DIFF
--- a/modules/common/engine/src/engine.ts
+++ b/modules/common/engine/src/engine.ts
@@ -49,7 +49,8 @@ export class CommonEngine {
         provide: INITIAL_CONFIG,
         useValue: {
           document: doc,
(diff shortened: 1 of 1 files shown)


alan-agius4:
@Dig-Doug, thanks for your contribution. However there is already an open PR for this.

See: https://github.com/angular/universal/pull/1860

PR #1881, opened 2020-10-30
fix: don't bundle common engine in hapi and express engine

alan-agius4:


CaerusKaru (commented):
Can you provide a short description of this change in the commit body? Otherwise LGTM

PR #1884, opened 2020-11-02
refactor: major clean up of all engines

alan-agius4:
- add return type to domContentLoadedFactory
- remove unused DI tokens
- Remove FileLoader implementations from all engines. Instead use the one in @nguniversal/common
- Extend RenderOptions Express and Hapi engines interfaces from Common Engine RenderOptions


modules/hapi-engine/README.md
```diff
@@ -87,7 +87,7 @@ You can access them by `@Inject`
 
 ```ts
 import { Request } from 'hapi';
-import { REQUEST } from '@nguniversal/hapi-engine/tokens';
+import { REQUEST } from '@nguniversal/hapi-engine';
```
CaerusKaru, line 90: Revert?

modules/common/tokens/index.ts
```diff
@@ -1,8 +0,0 @@
-/**
```
CaerusKaru, line 1: Is this technically a breaking change?

alan-agius4: They where marked with `ɵ`. Which indicated that these are private

modules/hapi-engine/README.md
alan-agius4: Done

CaerusKaru (approved):
Let's battle test this after we merge it to make sure nothing burned down, but LGTM. Nice work!

Issue #1505, opened 2020-02-13
npm run prerender does not create the ngsw-worker.js, ngsw.json files 


EdwinFdoGarcia:
# 🐞 Bug report

#### What modules are related to this issue?

- [ ] aspnetcore-engine
- [x] builders
- [ ] common
- [x] express-engine
- [ ] hapi-engine
- [ ] module-map-ngfactory-loader

### Is this a regression?
I don't know, possibly but before I used universal angular with module-map-ngfactory-loader and exported the files that I mentioned.

### Description
npm run prerender does not create the ngsw-worker.js, ngsw.json files.
create the application but only these two ngsw-worker.js, ngsw.json files would be missing.

## 🔬 Minimal Reproduction
``` 
npm run prerender

Hash: ef705016a62ee6d8132b
Time: 26852ms
Built at: 13/02/2020 12:02:42
  Asset      Size  Chunks                    Chunk Names
main.js  2.85 MiB       0  [emitted]  [big]  main
Entrypoint main [big] = main.js
chunk    {0} main.js (main) 5.85 MiB [entry] [rendered]
Generating ES5 bundles for differential loading...
ES5 bundle generation complete.

chunk {3} polyfills-es5.6e2f468dea03dd10e7eb.js, polyfills-es5.6e2f468dea03dd10e7eb.js.map (polyfills-es5) 127 kB [initial] [rendered]
chunk {2} polyfills-es2015.db39b23daa8851ccf344.js, polyfills-es2015.db39b23daa8851ccf344.js.map (polyfills) 35.7 kB [initial] [rendered]
chunk {1} main-es2015.dc112d754bfac23b83ec.js, main-es2015.dc112d754bfac23b83ec.js.map (main) 27.2 kB [initial] [rendered]
chunk {1} main-es5.dc112d754bfac23b83ec.js, main-es5.dc112d754bfac23b83ec.js.map (main) 27.3 kB [initial] [rendered]
chunk {0} runtime-es2015.0811dcefd377500b5b1a.js, runtime-es2015.0811dcefd377500b5b1a.js.map (runtime) 1.51 kB [entry] [rendered]
chunk {0} runtime-es5.0811dcefd377500b5b1a.js, runtime-es5.0811dcefd377500b5b1a.js.map (runtime) 1.51 kB [entry] [rendered]
[...]

PR #1887, opened 2020-11-05
feat(builders): support service worker generation with prerendering

alan-agius4:
Closes: #1505

### DIFF
--- a/modules/builders/src/prerender/index.ts
+++ b/modules/builders/src/prerender/index.ts
@@ -8,6 +8,9 @@
 
 import { BuilderContext, BuilderOutput, createBuilder, targetFromTargetString } from '@angular-devkit/architect';
 import { BrowserBuilderOptions } from '@angular-devkit/build-angular';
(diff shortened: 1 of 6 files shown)


modules/builders/src/prerender/index.ts
```diff
@@ -142,6 +158,26 @@ async function _renderUniversal(
...
+
+    if (browserOptions.serviceWorker) {
+      spinner.start('Generating service worker...');
+      try {
+        await augmentAppWithServiceWorker(
+          host,
+          root,
+          projectRoot,
+          normalize(outputPath),
+          browserOptions.baseHref || '/',
```
CaerusKaru, line 170: Do we want this in the case of explicitly blank href? Or should this be `??`?

CaerusKaru (approved):

modules/builders/src/prerender/index.ts
alan-agius4: I retained the same behaviour of the CLI. In many case you don't want to have a blank basehref because it will break routing

PR #1892, opened 2020-11-11
feat(fastify-engine): implement new package

### LINKED ISSUE #1871: Fastify Engine
# 🚀 Feature request

- [ ] aspnetcore-engine
- [ ] builders
- [ ] common
- [ ] express-engine
- [ ] hapi-engine
- [x] a new one

### Description
[@nestjs/ng-universal](https://github.com/nestjs/ng-universal) successively combine Angular, Angular Universal, NestJS and its features. The only exception is that NestJS is trying to strongly support Fastify what is not achieved in nestjs/ng-universal yet nestjs/ng-universal#3. What makes it even more important we can see that missing support for Fastify is slowing down the new features for Express powered apps nestjs/ng-universal#164. I have tried to add the implementation for Fastify, but as the implementation for Express uses an official package from angular/universal the custom one for Fastify can be considered as not stable enough.

There should be an official fastify-engine package.

Creating 3rd party packages with CommonEngine #1733

arturovt:
Closes #1871


alan-agius4:
Thanks for this PR and sorry for the late reply.

During our last team meeting we have discussed that instead of adding additional engines we should make the common engine API public. This will allow the community to create and publish Universal engine wrappers such as this one.

Thanks again.

PR #1914, opened 2020-12-08
feat: add hybrid rendering support

mgechev:
This change adds hybrid rendering support, which allows serving both:
- Server-side rendered pages
- Prerendered pages if they already exist

To avoid extra disk I/O operations, in both engine there's a
`pageExists` set that caches if the prerendered page exists. This way
on subsequent calls we'd be able to avoid the `existsSync` call, which
is on the critical path for serving SSR and prerendered pages.

Why not cache the entire page? Mostly to reduce memory consumption.
In apps with thousands of prerendered pages, this could cause server
crash. Also, in-memory cache might not be the most suitable for
autoscaling with multiple server instances.

### DIFF
--- a/integration/express-engine-ivy/server.ts
+++ b/integration/express-engine-ivy/server.ts
@@ -29,8 +29,21 @@ export function app(): express.Express {
     maxAge: '1y'
   }));
 
-  // All regular routes use the Universal engine
+  const pageExists = new Map<string, boolean>();
+  // All regular routes use the prerendered pages or the Universal engine
   server.get('*', (req, res) => {
+    const htmlPath = join(distFolder, req.path, 'index.html');
+    // Check if the page is already prerendered.
+    // To avoid extra disk I/O operation in the future remember if the file exists
+    if (!pageExists.has(req.path)) {
+      pageExists.set(req.path, existsSync(htmlPath));
+    }
+    if (pageExists.get(req.path)) {
+      res.sendFile(htmlPath);
+
+      return;
+    }
+    // Server-side rendering
     res.render(indexHtml, { req, providers: [{ provide: APP_BASE_HREF, useValue: req.baseUrl }] });
   });
 
--- a/integration/express-engine-ve/server.ts
+++ b/integration/express-engine-ve/server.ts
@@ -29,8 +29,21 @@ export function app(): express.Express {
     maxAge: '1y'
   }));
 
-  // All regular routes use the Universal engine
+  const pageExists = new Map<string, boolean>();
+  // All regular routes use the prerendered pages or the Universal engine
   server.get('*', (req, res) => {
+    const htmlPath = join(distFolder, req.path, 'index.html');
+    // Check if the page is already prerendered.
+    // To avoid extra disk I/O operation in the future remember if the file exists
+    if (!pageExists.has(req.path)) {
+      pageExists.set(req.path, existsSync(htmlPath));
+    }
+    if (pageExists.get(req.path)) {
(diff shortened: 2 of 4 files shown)


alan-agius4 (commented):
I spent sometime this morning thinking about our convo last night and how to address when the root page has been prerendering.

I think the below should do it;

```diff
import 'zone.js/dist/zone-node';

import { ngExpressEngine } from '@nguniversal/express-engine';
import * as express from 'express';
import { join } from 'path';

import { AppServerModule } from './src/main.server';
import { APP_BASE_HREF } from '@angular/common';
import { existsSync } from 'fs';

// The Express app is exported so that it can be used by serverless Functions.
export function app(): express.Express {
  const server = express();
  const distFolder = join(process.cwd(), 'dist/express-engine-ivy/browser');
+  const hasPrerenderedIndex = existsSync(join(distFolder, 'index.original.html'));
+  const indexHtml = hasPrerenderedIndex ? 'index.original.html' : 'index';
-  const indexHtml = existsSync(join(distFolder, 'index.original.html')) ? 'index.original.html' : 'index';

  // Our Universal express-engine (found @ https://github.com/angular/universal/tree/master/modules/express-engine)
  server.engine('html', ngExpressEngine({
    bootstrap: AppServerModule,
  }));

  server.set('view engine', 'html');
  server.set('views', distFolder);

  // Example Express Rest API endpoints
  // server.get('/api/**', (req, res) => { });
  // Serve static files from /browser
  server.get('*.*', express.static(distFolder, {
    maxAge: '1y'
  }));

  const pageExists = new Map<string, boolean>();
  // All regular routes use the prerendered pages or the Universal engine
  server.get('*', (req, res) => {
    const htmlPath = join(distFolder, req.path, 'index.html');
    // Check if the page is already prerendered.
    // To avoid extra disk I/O operation in the future remember if the file exists
    if (!pageExists.has(req.path)) {
+     if (req.path === '' && hasPrerenderedIndex) {
+       pageExists.set(req.path, false);
+     } else {
        pageExists.set(req.path, existsSync(htmlPath));
+     }
    }

    if (pageExists.get(req.path)) {
      res.sendFile(htmlPath);

      return;
    }

    // Server-side rendering
    res.render(indexHtml, { req, providers: [{ provide: APP_BASE_HREF, useValue: req.baseUrl }] });
  });

  return server;
}

function run() {
  const port = process.env.PORT || 4000;

  // Start up the Node server
  const server = app();
  server.listen(port, () => {
    console.log(`Node Express server listening on http://localhost:${port}`);
  });
}

// Webpack will replace 'require' with '__webpack_require__'
// '__non_webpack_require__' is a proxy to Node 'require'
// The below code is to ensure that the server is run only when not requiring the bundle.
declare const __non_webpack_require__: NodeRequire;
const mainModule = __non_webpack_require__.main;
const moduleFilename = mainModule && mainModule.filename || '';
if (moduleFilename === __filename || moduleFilename.includes('iisnode')) {
  run();
}

export * from './src/main.server';
```

I think in future this should be a seperate helper/middleware maybe to keep this logic hidden from end users.

I think with the changes for critical css, that it requires the browser directory to be known to the engine it can even be easier to achieve.

alan-agius4:
Note: the changes also need to be done in the actual schematics
- https://github.com/angular/universal/blob/master/modules/hapi-engine/schematics/install/files/__serverFileName%40stripTsExtension__.ts
- https://github.com/angular/universal/blob/master/modules/express-engine/schematics/install/files/__serverFileName%40stripTsExtension__.ts

PR #1917, opened 2020-12-11
feat(common): add hybrid rendering support

### LINKED ISSUE #1506: Make server.ts be able to serve prerendered files without having to write /index.html
# 🚀 Feature request

- [ ] aspnetcore-engine
- [x] builders
- [ ] common
- [ ] express-engine
- [ ] hapi-engine

### Description
I would like to make `server.ts` behave more like a CDN in that it would be able to serve prerendered files without having to go to `/[route]/index.html`.

I'd like to change the default installation of [`server.ts`](https://github.com/angular/universal/blob/master/modules/express-engine/schematics/install/files/__serverFileName%40stripTsExtension__.ts#L33) to do this:

```
  // If the given route has been prerendered, returns the file path to the cached html document.
  function getCachedFile(route) {
    const file = join(distFolder, route, 'index.html');
    if (existsSync(file)) {
      if (route !== '/') {
        return file;
      }
      const cachedIndexFile = join(distFolder, 'index.original.html');
      if (existsSync(cachedIndexFile)) {
        return cachedIndexFile;
      }
    }
  }

  // All regular routes use the Universal engine
  server.get('*', (req, res) => {
    const file = getCachedFile(req.path);
    if (file) {
      return res.sendFile(file);
    }
    res.render(indexHtml, { req, providers: [{ provide: APP_BASE_HREF, useValue: req.baseUrl }] });
  });
```

alan-agius4:
With this change we add hybrid rendering support in common engine, express engine and hapi engine.

Users don't need to do any changes to take advantage of this change.

The engine will serve the static prerendered page if a index.html exists for the given route, otherwise it will fallback to render the page in SSR.

Closes #1506


integration/express-engine-ivy-hybrid/package.json
```diff
@@ -11,7 +11,7 @@
     "ng": "ng",
     "start": "ng serve",
     "e2e": "ng e2e",
-    "prerender": "ng run express-engine-ivy-prerender:prerender --routes=/ --routes=pokemon/pikachu",
+    "prerender": "ng run express-engine-ivy-prerender:prerender --routes= --routes=pokemon/pikachu",
```
CaerusKaru, line 14: Do we need the first --routes?

integration/express-engine-ivy-hybrid/src/main.ts
```diff
@@ -8,7 +8,7 @@ if (environment.production) {
   enableProdMode();
 }
 
-document.addEventListener('DOMContentLoaded', () => {
+window as any)['doBootstrap'] = () => {
```
CaerusKaru, line 11: Is this missing a parens or is my code renderer going crazy?

modules/common/engine/src/engine.ts
```diff
@@ -65,6 +67,24 @@ export class CommonEngine {
...
+
+      if (pagePath !== resolve(opts.documentFilePath)) {
+        // View path doesn't match with prerender path.
+        let pageExists = this.pageExists.get(pagePath);
+        if (pageExists === undefined) {
+          pageExists = await exists(pagePath);
+          this.pageExists.set(pagePath, pageExists);
+        }
+
+        if (pageExists) {
```
CaerusKaru, line 81: Do we need this if?

alan-agius4: Yeah, because if it exists we want to send the static content and not render ssr.

integration/express-engine-ivy-hybrid/src/main.ts
alan-agius4: You are right, copy and paste mistake.

integration/express-engine-ivy-hybrid/package.json
alan-agius4: 👍

modules/common/engine/src/engine.ts
CaerusKaru: Then should the return type be string | undefined?

alan-agius4: It will always return a view. Either prerendered or SSR'd. As when page is not prerendered (IE: `pageExists  === false`) we fallback to use SSR.

CaerusKaru (approved):

modules/common/engine/src/engine.ts
alan-agius4: https://github.com/angular/universal/blob/256390da13efd1c3590449d43b4500a5c1d0ae1d/modules/common/engine/src/engine.ts#L69-L133

PR #1922, opened 2020-12-17
(FEAT) express-engine improved DX and SSL

SanderElias:
# Improved Dx for the express engine users

In the PR we adding 2 things.
1. SSL,  the universal app will now use SSL by default and create a self-signed certificate if none is provided.
2. there is a windows-shims-and-polyfill sample generated by the schematic.

# 2 shims and schematics.

This shows developers how they can add their own TS shims, Node shims, and node polyfills in a structured way, including the docs around this.


modules/express-engine/schematics/install/files/__serverFileName@stripTsExtension__.ts
```diff
@@ -3,11 +3,27 @@ import 'zone.js/dist/zone-node';
...
+import { AppServerModule } from './src/<%= stripTsExtension(main) %>';
+
+
+const serverSettings: ServerSettings = {
+  /** if you don't want to use SSL, switch this setting to false */
+  useSSL: true,
+  sslKey: '',
+  sslCert: '',
+  hostName: 'localHost',
+  port: Number.parseInt(process.env.PORT || '<%= serverPort %>' || '0', 10) || 4000,
```
alan-agius4, line 24: Why do we need all these new fallbacks? `serverPort` cannot be undefined hence anything after that is redundant.

```suggestion
  port: +process.env.PORT ?? <%= serverPort %>,
```

modules/express-engine/schematics/install/files/server-tools/add-ssl.ts
```diff
@@ -0,0 +1,71 @@
+import * as express from 'express';
```
alan-agius4, line 1: I have some concerns about adding self-ssl here, I think that such code should be added directly in the dev server and not in the users-code.

We should keep users code changes minimal. The more code there is, the harder it is for users to figure out things, and also it increases the complexity for us to add/improve on features because we'd need to do migrations.

The current implementation here will break dev-server since currently this only handles HTTP. https://github.com/angular/universal/blob/master/modules/builders/src/ssr-dev-server/index.ts

Adding SSL outside of dev-server might also slow drown incremental build, because on every incremental change the server is re-spawned at the moment and hence a certificate will need to be re-generated each time.

This also might cause a bit of confusion to users that now the dev-servers between universal SSR dev-server are different.

1) Universal dev-server is SSL by default (https://localhost:4200)
2) CLI is SSL opt-in (http://localhost:4200)

Does SSL during development add that much value?

modules/express-engine/schematics/install/files/server-tools/add-ssl.ts
```diff
@@ -0,0 +1,71 @@
...
+        name: 'angular-universal-test',
+        value: `${serverSettings.hostName}:${serverSettings.port}`,
+        type: 'RSAPublicKey',
+      },
+    ];
+    /**
+     * generate selfSigned here.
+     * If they remove the SSL there is no runtime penalty for loading this anymore.
+     */
+    const { generate } = __non_webpack_require__('selfsigned');
```
alan-agius4, line 49: Use an ES6 import here, and instead add `selfsigned` in `externalDepedencies`. https://github.com/angular/angular-cli/blob/faf7b26ef96880bec7e3f31891f3e8f02c418610/packages/angular_devkit/build_angular/src/server/schema.json#L213

Webpack specific syntax are considered unsupported, and unless there isn't any other way to achieve what we need we should not use Webpack syntax.

modules/express-engine/schematics/install/files/server-tools/add-ssl.ts
```diff
@@ -0,0 +1,71 @@
+import * as express from 'express';
+import { readFileSync } from 'fs';
+import * as https from 'https';
```
alan-agius4, line 3: Eventually when we there are no more comments left, we'd need to update all tests to reflect there changes https://github.com/angular/universal/tree/master/integration

Also, I am not seeing any changes to Hapi engine.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
```diff
@@ -0,0 +1,93 @@
+// tslint:disable: no-string-literal
```
alan-agius4, line 1: Why are these tslint comments needed here?

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
```diff
@@ -0,0 +1,93 @@
...
+// tslint:disable: typedef
+// tslint:disable: only-arrow-functions
+import { createWindow } from "domino";
```
alan-agius4, line 4: These will require domino to be added in the users project.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * overwrite existing window type for this polyfills file
+ * This is needed so we can create all the needed polyfills
+ * on Window without TS errors.
+ */
+interface Window {
```
alan-agius4, line 11:
```suggestion
type Window = Record<string, any>
```

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
```diff
@@ -0,0 +1,93 @@
...
+ * It can not be cancelled.
+ */
+global.requestAnimationFrame = window.requestAnimationFrame = function rqa(
+  callback: FrameRequestCallback
+) {
+  /**
+   * use a setTimeout to mimic requestAnimantionFrame
+   * it logs to console, so you can spot excessive use.
+   */
+  console.log(`requestAnimationFrame called`);
```
alan-agius4, line 81:
```suggestion
```

modules/express-engine/schematics/install/files/server-tools/server-settings.interface.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
alan-agius4, line 1: Maybe just move this in `add-ssl` file to reduce the files generated?

alan-agius4 (commented):
Note: I still haven't finished reviewing.

modules/express-engine/schematics/install/files/server-tools/add-ssl.ts
SanderElias: @alan-agius4 I did try the normal _import_ before I did this. (I only learned about this because it _is_ in the source of server.ts, to begin with!),  But the normal import didn't work. I'm willing to give this another go. But I think we need to discuss other issues first.

modules/express-engine/schematics/install/files/server-tools/add-ssl.ts
SanderElias: I was told to only concentrate on the express route.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
SanderElias: I'll double-check which ones are still needed.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
SanderElias: It is a full dependency of Universal anyway? The way current and near versions of yarn and npm work, it is going to be available anyway.  That's the pragmatic way of looking at it. We could add it to the users project, but I don't really see the added value there.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
SanderElias: Makes sense. will replace.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
SanderElias: I would strongly suggest leaving that in.  Some users might have recursive calls to RAF, this way they will have a pretty strong indicator on "why is my server stalled?" 
I thought of making a RAF replacement that only fires once, but that might cause other issues down the line that are much harder to debug

modules/express-engine/schematics/install/files/server-tools/server-settings.interface.ts
SanderElias: I just moved it out. Also, if we are going to move the AddSSL  out of userland, the interface is still valuable.
Perhaps we can move it along with the addSSL and then let users extend if they want to amend the interface

modules/express-engine/schematics/install/files/server-tools/add-ssl.ts
SanderElias: We have been discussing the exact same thing before. 
I also think the SSL support should go in the dev-server. We should discuss the details of this.
Also, I had an unresolved discussion issue about retaining the generated certificates in a `./cert` folder (that is a pretty common convention). As that would save on generation time.

alan-agius4 (commented):
Hi @SanderElias,

Let's schedule a call later on this week. 

I had a look at this again and I do have some concerns.

I am not fully grasping the main benefits of having https by default for development. This also is not fully implemented as currently the dev-server is yet to support HTTPS.

Such change would be a breaking change in terms of the dev-server, since switching the default might break existing users-workflow, in addition this might be confusing for existing users to why `ng serve` is http while `ng run app:serve-ssr` is https.

I would also expect that the `--ssl-key`, `--ssl-cert`, and `--ssl` option are part of the ssr-dev-server schema so that they are configurable the same way as `ng serve`.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
alan-agius4: I think that's not a good practice for a number of reasons.

1. There might also be the where 2 libraries dependent on 2 different versions of domino. In which case it would be unclear which version of the application is used.
2. You are relying on module hoisting, which does vary between packages
3. Domino can be removed from being a dependency of universal, and that shouldn't break existing projects.

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
```diff
@@ -0,0 +1,93 @@
...
+  /** Reference to the window. */
+  window: Window;
+  /** Reference to the document. */
+  document: Document;
+  /** Reference to the navigator. */
+  navigator: Navigator;
+  /** Reference to CSS. */
+  CSS: any;
+  /** Reference to Prism. */
+  Prism: any;
```
alan-agius4, line 31: Is this needed? How does this work in the browser context?

modules/express-engine/schematics/install/files/server-tools/windows-shims-and-polyfills.ts
```diff
@@ -0,0 +1,93 @@
...
+ * overwrite existing window type for this polyfills file
+ * This is needed so we can create all the needed polyfills
+ * on Window without TS errors.
+ */
+interface Window {
+  [prop: string]: any;
+  // Event:Event
+}
+
+const window: Window = createWindow("");
```
alan-agius4, line 16: Does providing an empty template cause any side-effects?

alan-agius4 (commented):
Hi @SanderElias,

Let's schedule a call later on this week. 

I had a look at this again and I do have some concerns.

## SSL 
I am not fully grasping the main benefits of having https by default for development. This also is not fully implemented as currently the dev-server is yet to support HTTPS.

Such change would be a breaking change in terms of the dev-server, since switching the default might break existing users-workflow, in addition this might be confusing for existing users to why `ng serve` is http while `ng run app:serve-ssr` is https.

I would also expect that the `--ssl-key`, `--ssl-cert`, and `--ssl` option are part of the ssr-dev-server schema so that they are configurable the same way as `ng serve`. 

## Shims

I have some doubts about this, this solves only the universal part, but does leave the problem for app-shell users. 

That said, there are a number of way shims can be done, even through 3rd packages, such as https://www.npmjs.com/package/@trilon/universal-ssr-mocks



Note for the future. it would be best to keep PRs scoped to a single feat/fix as it would be easier to land.

PR #2041, opened 2021-04-05
refactor: remove VE tests

alan-agius4:
In version 12, compilation of VE applications is longer supported.


CaerusKaru (approved):
I feel 2x lighter already

PR #2060, opened 2021-04-19
feat(@nguniversal/builders): add support for proxy configuration in ssr-dev-server

### LINKED ISSUE #1757: Universal support for proxy.conf.json
# 🚀 Feature request

- [ ] aspnetcore-engine
- [ ] builders
- [ ] common
- [x] express-engine
- [x] hapi-engine

### Description
Using Angular CLI, one is able to use a file (namely proxy.conf.json) and name this as proxyConfig in dev-server target options of angular.json, or as a parameter to ng serve.

It would be preferable to have proxyConfig as an available option for ssr-dev-server as well.

Three main workarounds spring to mind;
1) Routing middleware which redirects the request in the server.ts file
2) HTTP Interceptor
3) environment variable

However, all three requires changes to code. There is a reason the proxyConfig exists for dev-server; the same reasons could be applied for ssr-dev-server.

alan-agius4:
With this change we added a new `proxyConfig` option to be used to provide custom proxy configurations to the ssr-dev-server builder.

Closes #1757

### DIFF
--- a/modules/builders/src/ssr-dev-server/index.ts
+++ b/modules/builders/src/ssr-dev-server/index.ts
@@ -14,6 +14,7 @@ import {
 } from '@angular-devkit/architect';
 import { json, logging, tags } from '@angular-devkit/core';
 import * as browserSync from 'browser-sync';
+import { existsSync } from 'fs';
 import { createProxyMiddleware } from 'http-proxy-middleware';
 import { join, resolve as pathResolve } from 'path';
 import {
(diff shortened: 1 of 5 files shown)


modules/builders/src/ssr-dev-server/index.ts
```diff
@@ -322,4 +336,30 @@ function getSslConfig(
   return ssl;
 }
 
+function getProxyConfig(
+  root: string,
+  proxyConfig: string,
+): browserSync.MiddlewareHandler[] {
+  const proxyPath = pathResolve(root, proxyConfig);
+  if (!existsSync(proxyPath)) {
```
CaerusKaru, line 344: Didn't we realize we don't need this check, bc require will fail anyway if it doesn't exist?

alan-agius4: Yeah, although in this case it’s better to show a better error message, when file doesn’t exist.

Can use a try/catch block and filter out the `MODULE_NOT_FOUND`

alan-agius4: PTAL

CaerusKaru (approved):

PR #2073, opened 2021-04-28
feat(@nguniversal/builders): add static page generator builder using Angular SSR Clover

alan-agius4:
With this change we create a builder that can be used to generate static pages (prerendering) using the new Universal approach (Clover).

### DIFF
--- a/modules/builders/src/static-generator/index.ts
+++ b/modules/builders/src/static-generator/index.ts
@@ -1,0 +1,152 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {
+  BuilderContext,
+  BuilderOutput,
+  createBuilder,
+  targetFromTargetString,
+} from '@angular-devkit/architect';
+import { BrowserBuilderOptions, BrowserBuilderOutput } from '@angular-devkit/build-angular';
+import { normalizeOptimization } from '@angular-devkit/build-angular/src/utils/normalize-optimization';
+import { augmentAppWithServiceWorker } from '@angular-devkit/build-angular/src/utils/service-worker';
+import { normalize, resolve } from '@angular-devkit/core';
+import JestWorker from 'jest-worker';
+import * as ora from 'ora';
+import { cpus } from 'os';
+import * as path from 'path';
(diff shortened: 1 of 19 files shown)


modules/builders/src/static-generator/index.ts
```diff
@@ -0,0 +1,140 @@
...
+
+async function generateServiceWorker(
+  context: BuilderContext,
+  outputPath: string,
+  browserOptions: BrowserBuilderOptions,
+): Promise<BuilderOutput> {
+  const spinner = ora(`Generating service worker for ${outputPath}...`).start();
+
+  try {
+    const projectName = context.target && context.target.project;
```
kyliau, line 113:
```suggestion
    const projectName = context.target?.project;
```

modules/builders/src/static-generator/schema.ts
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Target to build.
+   */
+  browserTarget: string;
+  /**
+   * Whether or not the builder should extract routes and guess which paths to render.
+   */
+  guessRoutes?: boolean;
+  /**
+   * The number of cpus to use. Defaults to all but one.
```
kyliau, line 19: `all but one` -> `all minus one`?

modules/builders/src/static-generator/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Returns the union of routes, the contents of routesFile if given,
+ * and the static routes extracted if guessRoutes is set to true.
+ */
+export async function getRoutes(
+  options: PrerenderBuilderOptions,
+  tsConfigPath: string | undefined,
+  context: BuilderContext,
+): Promise<string[]> {
+  let routes = options.routes || [];
```
kyliau, line 25: nit: consider using `const` here, then `routes.push(...)` below

modules/builders/src/static-generator/worker.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { Engine } from '@nguniversal/common/clover/server';
+import { promises } from 'fs';
```
kyliau, line 10: nit: consider renaming `promises` to `fs`

kyliau (approved):

PR #2079, opened 2021-04-30
refactor: move schematic from common to express-engine

alan-agius4:
Previously, this schematic used to be shared and hence why it was in common. Now this is only used in express-engine.


modules/express-engine/schematics/install/index.ts
```diff
@@ -5,32 +5,294 @@
...
+  getProject,
+  stripTsExtension,
+} from '../utils';
+
+import { Schema as AddUniversalOptions } from './schema';
+
+const SERVE_SSR_TARGET_NAME = 'serve-ssr';
+const PRERENDER_TARGET_NAME = 'prerender';
+
+export function addUniversalCommonRule(options: AddUniversalOptions): Rule {
```
CaerusKaru, line 47: Do we still need this?

alan-agius4: No :)

CaerusKaru (approved):

Issue #2067, opened 2021-04-24
prerendering of 50k routes crashes on windows with ENAMETOOLONG


sierkov:
# 🐞 Bug report

#### What modules are related to this issue?
- [ ] aspnetcore-engine
- [x] builders
- [ ] common
- [ ] express-engine
- [ ] hapi-engine

### Is this a regression?
I think this was always broken.

### Description

npm run prerender fails with spawn ENAMETOOLONG error when rendering 50k routes on windows.
The root cause seems to be the OS limit on the command line length, which on windows is just 8000 bytes.
I believe other operating systems are affected too but require more routes to break.

In my opinion it's possible to fix this by adjusting _renderUniversal function in modules/builders/src/prerender/index.ts,
which at the moment takes all routes assigned for a prerender process and passes them as command line parameters.

I have provided a sample fix below, but I wasn't able to test it since I have never worked with bazel and couldn't rebuild @nguniversal with this patch.

## 🔬 Minimal Reproduction
1. ng new hugesite
2. cd hugesite
3. ng add @nguniversal/express-engine
4. add a route `  { path: 'some-long-enough-route-name-to-make-the-length-realistic/:xyz', component: AppComponent }`  to src/app/app-routing.module.ts
<pre>
import { NgModule } from '@angular/core';
import { RouterModule, Routes } from '@angular/router';
import { AppComponent } from './app.component';

const routes: Routes = [
  { path: 'some-long-enough-route-name-to-make-the-length-realistic/:xyz', component: AppComponent }
];

@NgModule({
  imports: [RouterModule.forRoot(routes, {
    initialNavigation: 'enabled'
})],
  exports: [RouterModule]
})
export class AppRoutingModule { }
</pre>
5. update "prerender.options" section in angular.json with ` "numProcesses": 1, "routesFile": "routes.txt"`
<pre>
"prerender": {
  "builder": "@nguniversal/builders:prerender",
  "options": {
    "browserTarget": "huge-project:build:production",
    "serverTarget": "huge-project:server:production",
    "routes": [],
    "numProcesses": 1,
    "routesFile": "routes.txt"
  },
  "configurations": {
    "production": {}
  }
}
</pre>
6. create mk-routes.js script to generate 50k random routes: 
<pre>
const fs = require('fs');

function genId() {
  const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
[...]

PR #2105, opened 2021-05-13
fix(@nguniversal/builders): make prerender work with large route lists

sierkov:
This is a fix for the issue described in: https://github.com/angular/universal/issues/2067


alan-agius4 (changes requested):
Why not simply read the index file from the worker?

In general, I'd also want to refactor this in the future and use jest-worker, like we do in https://github.com/angular/universal/blob/master/modules/builders/src/static-generator/index.ts

sierkov:
That part that passes the index file by value I have copied from the original code.
My best guess it is done that way because the author wanted all of the following at the same time:
1. add a comment "This page was prerendered with Angular Universal" into the output HTML file;
2. not to change the original source index file.
3. not to introduce an additional API parameter for the render, such as "buildComment", etc.
However, I'm new to the code base, so that's just IMHO.

Let me know you'd like to proceed. Should we change that part or keep it?

alan-agius4:
The additional of comment can be done directly in the worker, without adding an addition parameter.  The comment doesn't need to be passed on to the worker. 


I believe the intention here was to avoid additional FS reads, but now with your suggested change it is made obsolete.

sierkov:
I see. Thanks for the explanation. I have found one more dependency at line 113 of the original index.ts:
<pre>
if (normalizedStylesOptimization.inlineCritical) {
      // Workaround for https://github.com/GoogleChromeLabs/critters/issues/64
      indexHtml = indexHtml.replace(
        / media=\"print\" onload=\"this\.media='all'"><noscript><link .+?><\/noscript>/g,
        '>',
      );
}
</pre>

Would you like me to move all of the index.html processing to render.ts: adding the comment + normalizedStylesOptimization.inlineCritical?
I see that render.ts already does some postprocessing, so it would make sense to have it all in one place.

alan-agius4:
Yes please.

sierkov:
Done. LMK

modules/builders/src/prerender/index.ts
```diff
@@ -128,21 +117,23 @@ async function _renderUniversal(
...
+        (routesShard, shardIndex) =>
           new Promise((resolve, reject) => {
-            fork(workerFile, [
-              indexHtml.replace(
-                '</html>',
-                '<!-- This page was prerendered with Angular Universal -->\n</html>',
-              ),
+            const argsPath = path.join(
+              baseOutputPath,
+              `render-args-${process.pid}-${shardIndex}.json`,
```
alan-agius4, line 124: Do we still need this? consider we remove a large part of the content from the args by removing the index file? 

If so, we should really move and use jest-workers instead or at least as an interim until we move towards that we use an IPC channel to communicate with the child process. I am not very keen in creating this temporary file. As this feels like a workaround rather than a proper fix.

alan-agius4 (changes requested):
Thanks a lot for your contribution. Kindly see the feedback left.

modules/builders/src/prerender/index.ts
sierkov: I have not yet worked with jest-workers, so I am not able to comment on that immediately. As I understand, that looks like a simple and quick change to you, right?

Regarding the file and IPC, here are my thoughts:
1. My primary objective is to allow for much larger route lists: think 100k, 1M, or even 10M. That's where my current project is moving. In this case, even the reasonably large command-line-argument limit of Linux is not sufficient. So, replacing the file back with command line arguments is a no go.
2. Regarding other IPC methods (I assume one of shared memory, pipes, sockets, or SysV message passing), our case is more of initialization/configuration rather than a true IPC since we pass the arguments only one-way and do not need a response. A configuration file is a reasonable solution in this case event if it is a dynamic one. The only downside that I can see is the risk of the temporary file sticking when the parent process terminates abruptly. Since the file is stored in a temporary directory itself, IMHO, it's not a huge risk.

I understand where you are coming from - you need to maintain this codebase going forward, while I'm just trying to quickly fix one annoying issue. My concern is that the more changes I make in this patch, the more likely I to introduce some hard-to-see bug, since I'm new to this code base and can easily be not aware of some implicit requirements.

I can still take a look at jest-workers over next few days if you believe that will be a simple change. LMK

sierkov:
@alan-agius4, I'm waiting for your feedback regarding the next steps. JFYI

alan-agius4:
> @alan-agius4, I'm waiting for your feedback regarding the next steps. JFYI

I’ll try to get back to this in the coming days

modules/builders/src/prerender/index.ts
alan-agius4: Yeah, using jest-worker should be pretty straight forward and we are already using them in a similar fashion. See: https://github.com/angular/universal/blob/master/modules/builders/src/prerender/index.ts

I think this would be easier than using an IPC channel.

alan-agius4:
@sierkov, thanks for taking this. It’s much appreciated.

modules/builders/src/prerender/index.ts
alan-agius4: This would also make it easier for future contributes since we’ll use the same type of workers in all places.

modules/builders/src/prerender/worker.ts
```diff
@@ -0,0 +1,114 @@
...
+    };
+  }
+
+  if (renderModule && AppServerModule) {
+    // Happens when in Ivy mode.
+    return {
+      renderModuleFn: renderModule,
+      AppServerModuleDef: AppServerModule,
+    };
+  }
```
alan-agius4, line 109: We no longer need to support VE

```suggestion
    return {
      renderModuleFn: renderModule,
      AppServerModuleDef: AppServerModule,
    };
```

modules/builders/src/prerender/worker.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  throw new Error(
+    `renderModule method and/or AppServerModule were not exported from: ${bundlePath}.`,
+  );
+}
```
alan-agius4, line 114:
```suggestion
```

modules/builders/src/prerender/worker.ts
```diff
@@ -0,0 +1,114 @@
...
+export async function render(
+  outputPath: string,
+  serverBundlePath: string,
+  route: string,
+): Promise<void> {
+  const browserIndexOutputPath = path.join(outputPath, workerArgs.indexFile);
+  const outputFolderPath = path.join(outputPath, route);
+  const outputIndexPath = path.join(outputFolderPath, 'index.html');
+
+  const { renderModuleFn, AppServerModuleDef } = await getServerBundle(serverBundlePath);
```
alan-agius4, line 45:
```suggestion
  const { renderModule, AppServerModule } = await import (serverBundlePath);
```

alan-agius4 (approved):
Thanks for this, LGTM.
Just a couple of small comments.

Also, please squash the commits

sierkov:
@alan-agius4, thanks for the feedback. I'm mostly done with the squashing, but when I was re-reading through the patch a question arose:
The old code had the below lines to send errors and warnings received from the critical-css-inlining processor to the parent process. I am not quite sure how best to pass this information with jest-workers. Here is my current idea:
1. Print warnings to child's stderr and rely on the parent process for reading and reprinting the warnings (which it does). Lint has complained about my console prints due to no-console rule though. JFYI
2. Throw the first error returned by the processor as an exception. Some information will be lost but the remaining is still sufficient for a fix/rebuild loop to work.

The code lines of the original render.ts in question:
<pre>
    warnings?.forEach((message) => process.send!({ logLevel: 'warn', message }));
    errors?.forEach((message) => process.send!({ logLevel: 'error', message }));
</pre>

Do you have some suggestions regarding this situation?

Also, do you think we need to add a test or the already existing test set should be good enough?
I don't have ideas of simple but effective test cases at the moment, but I'd like to somehow increase my confidence that the new code works well in most use cases.

alan-agius4:
>     warnings?.forEach((message) => process.send!({ logLevel: 'warn', message }));
>     errors?.forEach((message) => process.send!({ logLevel: 'error', message }));

alan-agius4:
You can return an object literal with errors and warnings from the render function

Worker
```ts
export async function render(
  outputPath: string,
  serverBundlePath: string,
  route: string,
): Promise<{errors?: string[], warnings?:[]> {
}
```

Parent
```ts
const results = await Promise.all(
  routes.map((route) => (worker as any).render(outputPath, serverBundlePath, route)),
);


spinner.stop();
for (const {errors, warnings} of results) {
  errors?.forEach(e => context.logger.errror(e));
  warnings?.forEach(e => context.logger.warning(e));
}
spinner.start();
```

Regarding the test, I think in this case we can leave it out, since doing one will be non-trivial.

sierkov:
@alan-agius4, thanks! I've updated the patch and also made that in case of errors from css processor prerendering will be considered as failed. LMK

modules/builders/src/prerender/index.ts
```diff
@@ -103,93 +102,77 @@ async function _renderUniversal(
...
-          root,
-          projectRoot,
-          normalize(outputPath),
-          browserOptions.baseHref || '/',
-          browserOptions.ngswConfigPath,
-        );
+        const results = (await Promise.all(
+          routes.map((route) => (worker as any).render(outputPath, serverBundlePath, route)),
+        )) as RenderResult[];
+        const breakingErrors = [];
```
alan-agius4, line 136:
```suggestion
       let success = true;
```

modules/builders/src/prerender/index.ts
```diff
@@ -103,93 +102,77 @@ async function _renderUniversal(
...
+        for (const { errors, warnings } of results) {
+          errors?.forEach((e) => context.logger.error(e));
+          warnings?.forEach((e) => context.logger.warn(e));
+          if (errors) {
+            breakingErrors.push(...errors);
+          }
```
alan-agius4, line 142:
```suggestion
          success &&= errrors.length === 0
```

modules/builders/src/prerender/index.ts
```diff
@@ -103,93 +102,77 @@ async function _renderUniversal(
...
+        }
+        if (breakingErrors.length) {
+          throw Error(`Rendering finished with ${breakingErrors.length} errors.`);
+        }
```
alan-agius4, line 146:
```suggestion
        if (!success) {
          spinner.fail(`Prerendering routes to ${outputPath} failed. See above.`);
          return { success: false };
        }
```

sierkov: To reduce code duplication in error code paths? Any error within worker goes through the same catch block and to have one less spinner.fail and return call?

modules/builders/src/prerender/index.ts
sierkov: Would it work for you if I replace the array with a count? That way the error message will be slightly more informative - with the number of blocking issues?

alan-agius4: I am not too strongly about this, but I don't find that adding the count of errors adds anything actionable to the error message. It's up to you really.

modules/builders/src/prerender/index.ts
alan-agius4: Unless it's a fatal exception, I think it's better not to throw in this case.

It's make it harder to follow code as it's not intuitive without checking the whole code block. What happens with this `throw`? Does this cause a `success: false` or is a fatal exception?

sierkov: Regarding if it's fatal or not, that's a good question:
- This function uses both stop ASAP on error (parameters, server bundle check, service worker generation) and alway process all items (routes prerendering) error-handling strategies. Is there a reason for this?
- In the case, when this is a conscious decision (failed prerendering of some routes is not fatal), than we shouldn't be throwing that error at all. Otherwise, it should be OK to throw an error to indicate that the normal code path shall be interrupted.
Since this depends on the general technical design of the project, I need your input here.

Overall, the error handling in this function is complicated:
1. Both return codes and exceptions are used for error handling. 
2. Both stop-ASAP and keep-on-working error strategies are used.
3. The error information is used additionally to update the spinner's status.

There is a way to simplify it:
1. Use only exceptions inside the function and convert to return codes in the single outer try/catch/finally block.
2. Make spinners use error.message as the fail message instead of custom strings - spinner.fail is only called from catch blocks anyways.
3. Have only one try/catch block per function, so create two new smaller functions for routes prerendering and service worker generation.
However, this may take a few more iteration till we both like the end result. WDYT?

P.S. In the meantime, I updated the pull request to used an error counter and made the spinner calls in try/catch blocks consistent between routes generation and service-worker generation.

sierkov: Also, the latest CI integration test reports an integration failure during "npm run serve:ssr:dev exited with code SIGTERM" complaining about a time out - attached some output below. I cannot reproduce it locally - yarn test finishes successfully in a Docker container. I didn't change any files outside of prerendering, just rebased using the latest master.
Any suggestions out of the top of your head how to debug this?

<pre>
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] 
[1]   Hello world E2E Tests
[1]     ✗ should display: Hello world!
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1]     ✗ should populate window.location
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1] A Jasmine spec timed out. Resetting the WebDriver Control Flow.
[1]     ✗ stylesheets should be configured to load asynchronously
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
[1]       - Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
</pre>

alan-agius4: That looks like a flake.

alan-agius4: Overall, this logic can use a refactor, but the logic is as is, mostly because there are different types of errors.

- Exceptions
   When there is an exception we should exit the process because something when wrong and we don't know how to handle this error, where you might want to have the stack-trace in in the error log file etc...

- Diagnostic errors
   With this type of errors, we know how to handle these errors, we should collect as many diagnostic as possible before exiting so that the user can action these errors are quick as possible.

We should keep the distinction between an exception and a diagnostic error, and where possible the latter should be preferred.

modules/builders/src/prerender/index.ts
```diff
@@ -103,93 +102,75 @@ async function _renderUniversal(
...
-          browserOptions.baseHref || '/',
-          browserOptions.ngswConfigPath,
-        );
+        const results = (await Promise.all(
+          routes.map((route) => (worker as any).render(outputPath, serverBundlePath, route)),
+        )) as RenderResult[];
+        let numErrors = 0;
+        for (const { errors, warnings } of results) {
+          errors?.forEach((e) => context.logger.error(e));
+          warnings?.forEach((e) => context.logger.warn(e));
```
alan-agius4, line 139: Before printing anything, when the spinner is running you need to stop and start it, as otherwise it will overlap the  current spinner and cause broken output.


```suggestion
          spinner.stop();
          errors?.forEach((e) => context.logger.error(e));
          warnings?.forEach((e) => context.logger.warn(e));
           spinner.start();
```

alan-agius4 (approved):
Thanks again, one small last thing

modules/builders/src/prerender/index.ts
alan-agius4: Bump!

sierkov: Yep, it's on my list; plan to work on it this weekend.

modules/builders/src/prerender/index.ts
sierkov: Let's see if I understand you right:
1) Different types of errors - recoverable and non-recoverable (fatal) ones - makes sense. For example, C++ has two exception hierarchies: descendants of std:exception and of std:runtime_error to make it easy to differentiate between the two types of errors in the catch clauses. The only difference is that return codes instead of non-recoverable-exceptions are used. Right?
2) Diagnostic errors - this one can be a little too subtle for me. What's the difference between a warning and a diagnostic error? The closest analogy to diagnostic errors that comes to my mind is again from C++: the compiler collects and prints all compilation errors (won't stop at first one), but won't proceed to the linking phase if there was at least one compilation error. So there is a clear no-pass boundary till which the diagnostic errors are tolerated, but after which even a single error still leads to an abrupt termination of the process.

What's this no-pass boundary in this case? As the code is written right now, it is at service-worker generation: if there are prerendering issues, the service worker is not generated. Also, the next locale from browserResult.outputPaths won't be tried to be prerendered.
Is this fine?

P.S. I have incorporated your suggestion for the correct spinner updates and updated the pull request. LMK if there is any open issue with the code that I need to address.

alan-agius4 (approved):

PR #2117, opened 2021-05-20
fix(@nguniversal/common): add Window stubs to Engine (Clover)

### LINKED ISSUE #2108: [Clover] jsdom: Mocks for element scrolling methods are missing
I'm not sure if Clover is at a stage where you accept issues/PRs 😃 , so let me know if we should wait until you are over a certain point.

But one of the problems† I see is that some scrolling methods are not yet implemented in jsdom like `scrollTo`.
https://github.com/jsdom/jsdom/blob/master/lib/jsdom/browser/Window.js#L900

Which causes the server build to fail if a lib uses such a method.
```shell
Error: Not implemented: window.scrollTo
    at module.exports (<path/to/project>/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
    at <path/to/project>/node_modules/jsdom/lib/jsdom/browser/Window.js:866:7
    at rt.scrollToPosition (http://http/main.b51ee99a0a5badf2c7a5.js:1:111539)
    at c._next (http://http/main.b51ee99a0a5badf2c7a5.js:1:513321)
    at c.__tryOrUnsub (http://http/main.b51ee99a0a5badf2c7a5.js:1:1113718)
    at c.next (http://http/main.b51ee99a0a5badf2c7a5.js:1:1112947)
    at u._next (http://http/main.b51ee99a0a5badf2c7a5.js:1:1112126)
    at u.next (http://http/main.b51ee99a0a5badf2c7a5.js:1:1111900)
    at t.next (http://http/main.b51ee99a0a5badf2c7a5.js:1:1109640)
    at triggerEvent (http://http/main.b51ee99a0a5badf2c7a5.js:1:498683) undefined
 ```
 
 There is already an issue filled...
 https://github.com/jsdom/jsdom/issues/1422
 
 ...with an open PR:
 https://github.com/jsdom/jsdom/pull/2626
 
The question is: Should we mock it for Clover for now or wait and see? 
For example:
```ts
// server-engine.ts
 dom = new JSDOM(htmlContent, {
        runScripts: 'dangerously',
        resources: customResourceLoader,
        url: options.url,
        referrer: options.headers?.referrer as string | undefined,
        userAgent: options.headers?.['user-agent'] as string | undefined,
        beforeParse: (window) => {
          window.ngRenderMode = true;
          
          // TODO: remove it after https://github.com/jsdom/jsdom/pull/2626 is merged
          const noop = () => {};
          Object.defineProperty(window, 'scrollTo', { value: noop, writable: true });
          ...
        },
      });
  ```
 
† Another problem is routing.

## 🌍 Your Environment
```shell
Angular CLI: 12.0.0
Node: 12.18.0
Package Manager: yarn 1.22.10
OS: darwin x64

Angular: 12.0.0
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, material, platform-browser
... platform-browser-dynamic, platform-server, router

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1200.0
@angular-devkit/build-angular   12.0.0
@angular-devkit/core            12.0.0
@angular-devkit/schematics      12.0.0
@angular/fire                   6.1.4
@nguniversal/builders           12.0.0
@nguniversal/common             12.0.0
@schematics/angular             12.0.0
rxjs                            6.6.6
typescript                      4.2.4
```

alan-agius4:
Closes #2108

### DIFF
--- a/modules/common/clover/server/src/stubs.ts
+++ b/modules/common/clover/server/src/stubs.ts
@@ -1,0 +1,19 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { DOMWindow } from 'jsdom';
+
+function noop(): void {}
+
+export function augmentWindowWithStubs(window: DOMWindow): void {
+  window.resizeBy = noop;
+  window.resizeTo = noop;
+  window.scroll = noop;
+  window.scrollBy = noop;
+  window.scrollTo = noop;
+}
--- a/modules/common/clover/server/src/server-engine.ts
+++ b/modules/common/clover/server/src/server-engine.ts
@@ -16,6 +16,7 @@ import * as path from 'path';
 import { URL } from 'url';
 import { CustomResourceLoader } from './custom-resource-loader';
 import { InlineCriticalCssProcessor } from './inline-css-processor';
+import { augmentWindowWithStubs } from './stubs';
 
 export interface RenderOptions {
   headers?: Record<string, string | undefined | string[]>;
@@ -73,6 +74,7 @@ export class Engine {
         referrer: options.headers?.referrer as string | undefined,
         userAgent: options.headers?.['user-agent'] as string | undefined,
         beforeParse: (window) => {
+          augmentWindowWithStubs(window);
           window.ngRenderMode = true;
         },
       });


modules/common/clover/server/src/stubs.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { DOMWindow } from 'jsdom';
+
+function noop(): void {}
+
+export function augmentWindowWithStubs(window: DOMWindow): void {
+  window.resizeBy = noop;
```
CaerusKaru, line 14: Should we allow users to pass in their own versions? My instinct says there's no utility there but who knows. Approving without it, so up to you.

CaerusKaru (approved):

PR #2124, opened 2021-05-23
refactor(express-engine): use `projectName` for prerender

bampakoa:
Use the `projectName` variable in the prerender architect.

I used the `projectName` variable to be consistent with the `ssr-dev-server` architect.


alan-agius4 (approved):
LGTM, thanks

PR #2141, opened 2021-05-29
feat(@nguniversal/builders): spawn static server for build artifacts

### LINKED ISSUE #2132: [clover] Proposal: transferrable assets
# Proposal

#### What is the summary of the proposal?

This proposal adds a capability for static rendering with external content files.

Usecases

- Blog site
- Documentation
  - App Shell + Markdown contents (like AIO)
- Load initial data from JSON and fetch fresh one.

#### What is the proposal?

**Builder-side**

Adding a `transferAssets` option to the `@nguniversal/builders:static-generator`, which is an array that accepts paths of assets. It is relative from the output path of the browser target.

```json
       "build-static": {
          "builder": "@nguniversal/builders:static-generator",
          "options": {
            "browserTarget": "clover:build:production",
            "routes": [
              "/",
              "pokemon/pikachu",
              "pokemon/charmander",
              "pokemon/squirtle",
              "pokemon/bulbasaur"
            ],
            "transferAssets": [
              "pokemon.json"
            ]
          }
        }
```

**Engine-side**

Embedding assets as a JSON state immediately after `ngRenderMode` has been built. `TransferState` will be initialized with the state.

```ts
      // Embed assets within transfer state
      if (options.transferAssetsPaths) {
        const assets = await Promise.all(
          options.transferAssetsPaths.map((assetPath) => ({
            assetPath,
            content: fs.promises.readFile(path.resolve(options.publicPath, assetPath), 'utf-8'),
          })),
        );
        const assetMap = assets.reduce(
          (map, asset) => ({ ...map, [asset.assetPath]: asset.content }),
          {},
        );
        const state = JSON.stringify(assetMap);
        this.embedTransferState(doc, ngRenderMode.appId, state);
      }

      await ngRenderMode.getWhenStable();
```

This is an implementation for proof of concept. I don't think it is production-ready but enough to explain what this proposal aims for.

https://github.com/lacolaco/universal/commit/52f7b7de5099bf0bc9fccbc8a914e92374572ca2

#### Is there anything else we should know?

**Considerations**

- jsdom doesn't support intercepting to XHR, so `CustomResourceLoader` cannot help this usecase. 
- At first, I thought `assets` from the browser target can be used for this purpose, but assets can contain non-text files like images. So explicit option would be needed.

lacolaco:
Using express to spawn a local static server for browser build artifacts which allows HTTP requests to a relative path like `/assets/data.json` while rendering.

Close #2132 

cc/ @alan-agius4

### DIFF
--- a/integration/clover/test.js
+++ b/integration/clover/test.js
@@ -21,4 +21,12 @@ for (const page of pages) {
   if (!content.includes('ng-clover')) {
     throw new Error(`Page ${page} didn't contain ng-clover marker.`);
   }
+
+  // Asserts embedded state transfer for local relative requests
+  if (page === 'index.html') {
+    const dataJson = require('./src/assets/data.json');
+    if (!content.includes(dataJson.message)) {
+      throw new Error(`Page ${page} didn't contain message from data.json.`);
+    }
+  }
 }
--- a/modules/builders/package.json
+++ b/modules/builders/package.json
@@ -32,6 +32,7 @@
     "@angular-devkit/core": "DEVKIT_CORE_VERSION",
     "@nguniversal/common": "0.0.0-PLACEHOLDER",
     "browser-sync": "^2.26.7",
+    "express": "^4.17.1",
     "guess-parser": "^0.4.12",
     "http-proxy-middleware": "^2.0.0",
     "jest-worker": "26.6.2",
--- a/modules/builders/src/static-generator/index.ts
+++ b/modules/builders/src/static-generator/index.ts
@@ -16,10 +16,14 @@ import { BrowserBuilderOptions, BrowserBuilderOutput } from '@angular-devkit/bui
 import { normalizeOptimization } from '@angular-devkit/build-angular/src/utils/normalize-optimization';
 import { augmentAppWithServiceWorker } from '@angular-devkit/build-angular/src/utils/service-worker';
 import { normalize, resolve } from '@angular-devkit/core';
+import * as express from 'express';
+import * as http from 'http';
 import JestWorker from 'jest-worker';
 import * as ora from 'ora';
 import { cpus } from 'os';
 import * as path from 'path';
+import { promisify } from 'util';
+import { getAvailablePort } from '../ssr-dev-server/utils';
 import { Schema as PrerenderBuilderOptions } from './schema';
 import { getRoutes } from './utils';
 import { WorkerSetupArgs } from './worker';
@@ -34,9 +38,9 @@ export async function execute(
   context: BuilderContext,
 ): Promise<BuilderOutput> {
   const browserTarget = targetFromTargetString(options.browserTarget);
-  const browserOptions = ((await context.getTargetOptions(
+  const browserOptions = (await context.getTargetOptions(
     browserTarget,
-  )) as unknown) as BrowserBuilderOptions;
+  )) as unknown as BrowserBuilderOptions;
   const routes = await getRoutes(options, browserOptions.tsConfig, context);
 
   if (!routes.length) {
@@ -57,12 +61,15 @@ export async function execute(
   try {
     for (const outputPath of outputPaths) {
       const spinner = ora(`Prerendering ${routes.length} route(s) to ${outputPath}...`).start();
+
+      const staticServer = await createStaticServer(outputPath);
       try {
         await Promise.all(
           routes.map((route) =>
             (worker as any).render({
               outputPath,
               route,
+              port: staticServer.port,
             }),
           ),
         );
@@ -79,6 +86,8 @@ export async function execute(
         spinner.fail(`Prerendering routes to ${outputPath} failed.`);
 
         return { success: false, error: error.message };
+      } finally {
+        await staticServer.close();
       }
     }
 
@@ -88,6 +97,22 @@ export async function execute(
     // tslint:disable-next-line: no-floating-promises
     const _ = worker.end();
   }
+}
+
+async function createStaticServer(browserOutputRoot: string): Promise<{
+  close: () => Promise<void>;
+  port: number;
+}> {
+  const app = express();
+  app.use(express.static(browserOutputRoot));
+  const port = await getAvailablePort();
+  const server = new http.Server(app);
+  await new Promise<void>((res) => server.listen(port, res));
+
+  return {
+    close: promisify(server.close.bind(server)),
+    port,
+  };
 }
 
 function createWorker(browserOptions: BrowserBuilderOptions): JestWorker {
--- a/integration/clover/src/app/app.module.ts
+++ b/integration/clover/src/app/app.module.ts
@@ -20,7 +20,10 @@ import { PokedexComponent } from './pokedex.component';
     // The HttpClientInMemoryWebApiModule module intercepts HTTP requests
     // and returns simulated server responses.
 
-    HttpClientInMemoryWebApiModule.forRoot(PokemonService, { dataEncapsulation: false }),
+    HttpClientInMemoryWebApiModule.forRoot(PokemonService, {
+      dataEncapsulation: false,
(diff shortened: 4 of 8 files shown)


modules/builders/src/static-generator/index.ts
```diff
@@ -57,12 +59,18 @@ export async function execute(
   try {
     for (const outputPath of outputPaths) {
       const spinner = ora(`Prerendering ${routes.length} route(s) to ${outputPath}...`).start();
+
+      const port = await getAvailablePort();
+      const staticServer = createStaticServer(context, outputPath).listen(port, () => {
```
alan-agius4, line 64: This might cause a race condition. The server is spawning is async, which means that we might start processing the requests prior of the server being ready to take requests.

Also, the port assigned can be moved in the createStaticServer method.
```ts
async function createStaticServer(browserOutputRoot: string): {
    app: express.Express;
    port: number;
} {
    const server = express();
    app.use(express.static(browserOutputRoot));

    const port = await getAvailablePort();
    await promisify(app.listen)(port);

    return {
        app,
        port,
    };
  }
```

modules/builders/src/static-generator/index.ts
```diff
@@ -90,6 +100,17 @@ export async function execute(
   }
 }
 
+function createStaticServer(context: BuilderContext, browserOutputRoot: string): express.Express {
+  const app = express();
+  app.use(express.static(browserOutputRoot));
+  app.use((req, res, next) => {
+    context.logger.debug(`[static-server] ${res.statusCode}: ${req.method} ${req.originalUrl}`);
```
alan-agius4, line 107: Why do we need a fallback middleware to add logs? Express already has it's own debugging loggings

`DEBUG = express:*`

modules/builders/src/static-generator/index.ts
```diff
@@ -57,12 +59,18 @@ export async function execute(
   try {
     for (const outputPath of outputPaths) {
       const spinner = ora(`Prerendering ${routes.length} route(s) to ${outputPath}...`).start();
+
+      const port = await getAvailablePort();
+      const staticServer = createStaticServer(context, outputPath).listen(port, () => {
+        spinner.info(`Local static server is running on localhost:${port}`);
```
alan-agius4, line 65: I think the info here, isn't needed, as the consumers shouldn't really care which port is used.

modules/builders/src/prerender/index.ts
```diff
@@ -59,8 +59,8 @@ async function _scheduleBuilds(
 
   try {
     const [browserResult, serverResult] = await Promise.all([
-      (browserTargetRun.result as unknown) as BuildBuilderOutput,
-      (serverTargetRun.result as unknown) as BuildBuilderOutput,
+      browserTargetRun.result as unknown as BuildBuilderOutput,
```
alan-agius4, line 62: Please revert, to keep the PR focused.

modules/builders/src/prerender/index.ts
```diff
@@ -205,9 +205,9 @@ export async function execute(
   context: BuilderContext,
 ): Promise<PrerenderBuilderOutput> {
   const browserTarget = targetFromTargetString(options.browserTarget);
-  const browserOptions = ((await context.getTargetOptions(
+  const browserOptions = (await context.getTargetOptions(
```
alan-agius4, line 208: Please revert, to keep the PR focused.

modules/builders/src/prerender/utils.spec.ts
```diff
@@ -28,10 +28,10 @@ describe('Prerender Builder Utils', () => {
     ];
 
     const TSCONFIG_PATH = 'tsconfig.app.json';
-    const CONTEXT = ({
+    const CONTEXT = {
```
alan-agius4, line 31: Please revert, to keep the PR focused.

modules/builders/src/prerender/render.ts
```diff
@@ -25,12 +25,8 @@ const [
  * Handles importing the server bundle.
  */
 async function getServerBundle(bundlePath: string) {
-  const {
-    AppServerModule,
-    AppServerModuleNgFactory,
-    renderModule,
-    renderModuleFactory,
-  } = await import(bundlePath);
+  const { AppServerModule, AppServerModuleNgFactory, renderModule, renderModuleFactory } =
```
alan-agius4, line 28: Please revert, to keep the PR focused.

modules/builders/package.json
```diff
@@ -32,6 +32,7 @@
     "@angular-devkit/core": "DEVKIT_CORE_VERSION",
     "@nguniversal/common": "0.0.0-PLACEHOLDER",
     "browser-sync": "^2.26.7",
+    "express": "EXPRESS_VERSION",
```
alan-agius4, line 35: In here, you can use the actual version, no need to use the replacements. 
```suggestion
    "express": "^4.17.1",
```

alan-agius4:
Can you please add a basic test in https://github.com/angular/universal/blob/master/integration/clover/test.js?

alan-agius4 (commented):
Couple of small NITS.

Thanks for this @lacolaco.

lacolaco:
@alan-agius4 Thank you for the review. I've fixed all that was pointed out. BTW, unfortunately, `promisify(app.listen)` that you suggested doesn't work as expected.

modules/builders/src/static-generator/index.ts
```diff
@@ -79,6 +85,8 @@ export async function execute(
         spinner.fail(`Prerendering routes to ${outputPath} failed.`);
 
         return { success: false, error: error.message };
+      } finally {
+        staticServer.close();
```
alan-agius4, line 89: Let's make sure to close the server prior to starting a new one. This is not an async call. 
Let's turn this into a promise.

```suggestion
       await staticServer.close();
```

modules/builders/src/static-generator/index.ts
```diff
@@ -90,6 +98,21 @@ export async function execute(
...
+  close: Function;
+  port: number;
+}> {
+  const app = express();
+  app.use(express.static(browserOutputRoot));
+  const port = await getAvailablePort();
+  const server = new http.Server(app);
+  await new Promise<void>((resolve) => server.listen(port, resolve));
+  return {
+    close: server.close.bind(server),
```
alan-agius4, line 111:
```suggestion
    close: promisify(server.close),
```

modules/builders/src/static-generator/index.ts
```diff
@@ -90,6 +98,21 @@ export async function execute(
   }
 }
 
+async function createStaticServer(browserOutputRoot: string): Promise<{
+  close: Function;
```
alan-agius4, line 102:
```suggestion
  close: () => Promise<void>;
```

integration/clover/test.js
```diff
@@ -21,4 +21,12 @@ for (const page of pages) {
   if (!content.includes('ng-clover')) {
     throw new Error(`Page ${page} didn't contain ng-clover marker.`);
   }
+
+  // Asserts embeded JSON content
```
alan-agius4, line 25:
```suggestion
  // Asserts embedded state transfer for local relative requests
```

alan-agius4 (approved):
Last couple of things.

Can you also please squash? Thanks.

modules/builders/src/static-generator/index.ts
lacolaco: `bind` is necessary because that calls `this` internally. I'll add it. https://github.com/nodejs/node/issues/21482

lacolaco:
CI test fails due to unrelated files. Is it a flaky test?

alan-agius4:
> CI test fails due to unrelated files. Is it a flaky test?

Looks like it.

alan-agius4 (approved):
Thanks again.

PR #2192, opened 2021-07-03
fix(@nguniversal/common-engine):  `inlineCriticalCss` defaults to false

ikeq:
https://github.com/angular/universal/blob/master/modules/common/engine/src/engine.ts#L31


ikeq:
@googlebot I signed it!

alan-agius4 (changes requested):
Thanks for your contribution, by the default of true is correct.

Can you update the typedoc?

For this change the commit message should be

```
docs(@nguniversal/common): inlineCriticalCss defaults to true
```

alan-agius4 (approved):
Thanks

PR #2235, opened 2021-08-09
ci: replace Keen with Doug in GitHub CODEOWNERS

kyliau:


alan-agius4:
Can you also please update https://github.com/angular/universal/blob/master/README.md?

alan-agius4 (approved):

PR #2250, opened 2021-08-18
fix: remove the toPromise()

HyperLife1119:

PR #2251, opened 2021-08-18
Revert "fix: remove the toPromise()"

alan-agius4:
Reverts angular/universal#2250

Following a convo with @CaerusKaru 

Promise is  there because Promises are microtasks vs Observables which are macrotasks. And that isStable only fires when the macrotask queue is empty, which it would never be since that last observable wraps it


CaerusKaru (approved):

CaerusKaru:
Blocking while we investigate this a little further.

PR #2345, opened 2021-10-05
build: ship partial compilation bundles complaint with APF v13

alan-agius4:


modules/builders/src/prerender/utils.spec.ts
```diff
@@ -15,7 +15,7 @@ import * as guessParser from 'guess-parser';
 import { PrerenderBuilderOptions } from './models';
 import { getIndexOutputFile, getRoutes, shardArray } from './utils';
 
-describe('Prerender Builder Utils', () => {
+xdescribe('Prerender Builder Utils', () => {
```
alan-agius4, line 18: These tests are not ESM complaint.

.bazelrc
```diff
@@ -16,7 +16,7 @@ test:debug --test_arg=--node_options=--inspect-brk --test_output=streamed --test
 build --symlink_prefix=dist/
 
 # Performance: avoid stat'ing input files
-build --watchfs
+build --nowatchfs
```
CaerusKaru, line 19: If we're changing this should we also change the comment?

integration/express-engine-ivy-hybrid/package.json
```diff
@@ -36,7 +36,7 @@
     "@nguniversal/express-engine": "file:../../dist/modules-dist/express-engine",
     "angular-in-memory-web-api": "^0.11.0",
     "express": "file:../../node_modules/express",
-    "rxjs": "file:../../node_modules/rxjs",
+    "rxjs": "7.3.1",
```
CaerusKaru, line 39: Why was this changed?

modules/builders/src/utils/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import { URL } from 'url';
+
+/**
+ * This uses a dynamic import to load a module which may be ESM.
```
CaerusKaru, line 12: Should we add a TODO to readdress this with TS 4.5?

CaerusKaru (approved):

PR #2366, opened 2021-10-11
refactor(@nguniversal/builders): replace `jest-worker` with `piscina`

alan-agius4:
We no longer need to support old Node.JS versions which require fallback to forked proceses. Therefore, we replace the `jest-worker` with `piscina` which is a lighter weight worker pool.

Side note: `piscina` is also developed by one of the Node.JS core maintainers.

### DIFF
--- a/modules/bazel-tsconfig-build.json
+++ b/modules/bazel-tsconfig-build.json
@@ -20,7 +20,14 @@
     "lib": ["es2018", "dom"],
     "skipLibCheck": true,
     "types": ["tslib", "node"],
-    "typeRoots": ["node_modules/@types"]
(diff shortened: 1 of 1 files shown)


modules/builders/src/prerender/worker.ts
```diff
@@ -10,49 +10,49 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { loadEsmModule } from '../utils/utils';
 
-export interface WorkerSetupArgs {
+export interface RenderOptions {
```
CaerusKaru, line 13: Is this a breaking change?

PR #2426, opened 2021-11-03
docs: update `CHANGELOG` for `v13.0.0`

dgp1130:
Also adds `12.1.2` and `12.1.3` release notes which apparently got missed previously. They should align with the existing GitHub releases.

/cc @alan-agius4 for sanity check, but due to timezone differences we'll merge this ASAP and just any issues afterwards.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,32 +1,53 @@
-<a name="13.0.0-rc.0"></a>
+<a name="13.0.0"></a>
 
-# 13.0.0-rc.0 (2021-10-14)
-
-## Breaking Changes
-
-### @nguniversal/common
-
-- Inlining of critical CSS is no longer enable by default. Users already on Angular version 12 and have not opted-out from using this feature are encouraged to opt-in using the `inlineCriticalCss` option.
-
-The motivation behind this change is that the package used to parse the CSS has a number of defects which can lead to unactionable error messages when updating to Angular 13 from versions priors to 12. Such errors can be seen in the following issue https://github.com/angular/angular-cli/issues/20760.
-
-### @nguniversal/builders
-
-| Commit                                                                                           | Type | Description                                  |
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

dgp1130:
These are mostly auto-generated but I removed the note about inline critical CSS being disabled because that change was later reverted, so there is no change to users.

alan-agius4 (approved):

CHANGELOG.md
```diff
@@ -1,32 +1,59 @@
...
-# 13.0.0-rc.0 (2021-10-14)
+# 13.0.0 (2021-11-03)
 
 ## Breaking Changes
 
 ### @nguniversal/common
 
-- Inlining of critical CSS is no longer enable by default. Users already on Angular version 12 and have not opted-out from using this feature are encouraged to opt-in using the `inlineCriticalCss` option.
-
 The motivation behind this change is that the package used to parse the CSS has a number of defects which can lead to unactionable error messages when updating to Angular 13 from versions priors to 12. Such errors can be seen in the following issue https://github.com/angular/angular-cli/issues/20760.
```
alan-agius4, line 9:
```suggestion
The motivation behind this change is that the package used to parse the CSS has a number of defects which can lead to unactionable error messages when updating to Angular 13 from versions priors to 12. Such errors can be seen in the following issue https://github.com/angular/angular-cli/issues/20760.
```

alan-agius4 (changes requested):
Actually one minor change

CHANGELOG.md
dgp1130: Ah yes, good catch. Thanks Alan!

alan-agius4 (approved):

dgp1130:
Also cherry picked to `13.0.x` in https://github.com/angular/universal/commit/d0199f1fbc86631629ecdb3254c32b62b9ab95c6.

PR #2427, opened 2021-11-03
release: v13.1.0-next.0

dgp1130:
Updates Angular deps to `v13.1.0-next.0`. The package version is already `13.1.0-next.0`, so nothing to update there.

This keeps the `next` NPM tag pointing to a different release than `latest`. Unfortunately I can't push directly to `master` so a PR is required.


josephperrott (approved):
LGTM

PR #2428, opened 2021-11-03
docs: update `CHANGELOG` with `v13.1.0-next.0` release notes

dgp1130:
This is the same as `v13.0.0`, just with a new version number so `next` and `latest` tag different versions in NPM.


josephperrott (approved):
LGTM

PR #2447, opened 2021-11-17
build: update ng-dev configuration for release

alan-agius4:
Update the configuration for releasing via ng-dev.

```
yarn ng-dev release publish
```

PR #2450, opened 2021-11-19
docs: update release docs to reflect new tooling

dgp1130:
Release tooling was updated in #2447 to use dev-infra's unified system. The actual documentation was mostly copied from the existing [Angular CLI release docs](https://github.com/angular/angular-cli/blob/e0a1a94e79a3b0a5ec03bb41c70e8f8674557c10/docs/process/release.md), with the key difference being that releases and published as necessary rather than every week.

### DIFF
--- a/docs/process/release.md
+++ b/docs/process/release.md
@@ -1,65 +1,10 @@
-# Release Steps
+# Release
 
-- Login as `angular` npm account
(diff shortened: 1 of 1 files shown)


josephperrott (approved):
LGTM

alan-agius4 (approved):
Thanks @dgp1130.

PR #2477, opened 2021-12-13
# Feature - Optional waiting time before Reloading Chrome

conradkirschner:
Gives Developers the option to set a waiting time before reloading the application.
This is for reducing stress on slower Developer devices.


hiepxanh:
@alan-agius4 what do you think?
 I think developer should upgrade the device. Sorry to say that, but I still have no idea why we need to wait? is that I should load faster? <3

conradkirschner:
@hiepxanh I think there is a high chance that we have a bug in our application that cause this issue. 

Issue: on code change the application gets rebuild and triggers an refresh. Browser is stuck in refresh. 
Resolution is that I added some waiting there, then refresh works as expected. 
OR cancel request, and trigger refresh manually.

Somewhere is something wrong, but I don't see where... also no additional logging, what isn't finished. (it says it's finished but then get stucked) 

totally fine with dropping this PR, since timeouts are bad and this should be fixed somehow differently.

alan-agius4:
Closing as per https://github.com/angular/universal/pull/2477#issuecomment-1062936904

PR #2507, opened 2022-01-14
CI Fixes

alan-agius4:


CaerusKaru:
With dependency changes, should this be targeting major?

alan-agius4:
> With dependency changes, should this be targeting major?

Discussed offline.

PR #2570, opened 2022-03-08
fix(express-engine): construct req url instead of using originalUrl

### LINKED ISSUE #2564: Please use req.baseUrl + req.url instead of req.originalUrl
Hi,
i think this usage of `req.originalUrl` is a mistake:
https://github.com/angular/universal/blob/1adb89372704be7761e5c2ac8efeda918342ac0b/modules/express-engine/src/main.ts#L51

Its goal is to not get the mounted path: to achieve that goal, the right way is to use `${req.baseUrl}${req.url}`.

Some other middlewares might be changing req.url (lawfully) and using `originalUrl` shunts them.
See https://github.com/kapouer/express-urlrewrite/pull/1 and years later https://github.com/kapouer/express-urlrewrite/issues/8.

CaerusKaru:
The `originalUrl` as part of the request is meant to be immutable,
but some users may want the URL to be mutated by middleware. By
locking our URL construction to the originalUrl, we rob users of
this ability. Therefore, we construct the URL using comparable
elements of the request object.

Fixes #2564

### DIFF
--- a/modules/express-engine/src/main.ts
+++ b/modules/express-engine/src/main.ts
@@ -6,10 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import { StaticProvider } from '@angular/core';
+import type { StaticProvider } from '@angular/core';
 import { CommonEngine, RenderOptions as CommonRenderOptions } from '@nguniversal/common/engine';
 import { REQUEST, RESPONSE } from '@nguniversal/express-engine/tokens';
-import { Request, Response } from 'express';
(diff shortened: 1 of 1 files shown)


CaerusKaru:
@alan-agius4 I'm happy to mark this as a breaking change and update the commit message. My thinking on not: we don't explicitly guarantee this behavior as part of the contract with users. And yet I still think it's "breaking" enough that it should be in a major version bump.

alan-agius4 (approved):
Targeting major sounds reasonable to me.

PR #2580, opened 2022-03-15
feat(express-engine): schematics compatible with Node on Apache Passenger

distante:
When deploying Universal to a Shared Hosting with Apache Passenger the server never starts because the `if` block is always false.

This PR adds an extra `of` checking if we are on a Passenger Env by checking if `moduleFilename` includes the path to [lsnode](https://github.com/rperper/lsnode).

Also starts the server without a port because no port is needed on Apache Passenger.

PS: I wanted to create a test for this but I could not figure out where...

### DIFF
--- a/integration/common-engine/server.ts
+++ b/integration/common-engine/server.ts
@@ -42,11 +42,24 @@ export function app(): express.Express {
   return server;
 }
 
-function run() {
-  const port = process.env.PORT || 4200;
+// on Passenger, the moduleFilename will contain a/path/to/lsnode.js
+function isRunningOnApachePassenger(): boolean {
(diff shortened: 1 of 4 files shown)


distante:
I'm not sure where should I look at the failing tests :/

distante:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> For more information, open the [CLA check for this pull request](https://github.com/angular/universal/pull/2580/checks?check_run_id=5560893489).

Already signed.

metal-draken:
will this be merged soon?

PR #2623, opened 2022-04-25
build: changes for primary branch rename

devversion:
Changes from the `DIRECT` phase of the renaming master
to main planning doc.

Note: Skipping LTS backports here since we don't make any guarantees for the universal repo
and its likely not worth (we would need more ng-dev backports there as well)


devversion:
cc. @josephperrott (don't have access to this repo)

alan-agius4 (approved):

PR #2666, opened 2022-05-24
fix(@nguniversal/common): add a marker for SSR pages

### LINKED ISSUE #2612: Add a marker for SSR pages
# 🚀 Feature request

Add a marker for SSR pages.

### Description

We'd want to easily detect which pages are server-side rendered in HTTPArchive.

Add a marker in an attribute.

Currently, we can easier distinguish between SSR, SSG, and client-side apps (CSR) by:
- SSG adds a comment saying the page is prerendered
- SSR output looks different than CSR apps but it's harder to query in HTTPArchive which seem to primarily allow queries over the final output, after hydration

// cc @kara @jelbourn

alan-agius4:
This provides an easy way to distinguish between SSR, SSG, and client-side apps (CSR).

Closes #2612

//cc @mgechev

### DIFF
--- a/modules/builders/src/prerender/worker.ts
+++ b/modules/builders/src/prerender/worker.ts
@@ -50,7 +50,7 @@ export async function render({
   let indexHtml = await fs.promises.readFile(browserIndexInputPath, 'utf8');
   indexHtml = indexHtml.replace(
     '</html>',
-    '<!-- This page was prerendered with Angular Universal -->\n</html>',
+    '<!-- This page was prerendered with Angular Universal (SSG) -->\n</html>',
   );
   if (inlineCriticalCss) {
     // Workaround for https://github.com/GoogleChromeLabs/critters/issues/64
--- a/modules/common/engine/src/engine.ts
+++ b/modules/common/engine/src/engine.ts
@@ -101,7 +101,11 @@ export class CommonEngine {
     }
 
     const moduleOrFactory = this.module || opts.bootstrap;
-    const html = await renderModule(moduleOrFactory, { extraProviders });
+    const html = (await renderModule(moduleOrFactory, { extraProviders })).replace(
+      '</html>',
+      '<!-- This page was rendered with Angular Universal (SSR) -->\n</html>',
+    );
+
     if (!inlineCriticalCss) {
       return html;
     }


alan-agius4:
`platform-server` doesn't know in which context the module is being rendered and I don't think it should. It shouldn't matter if we are rendering an app-shell, SSG or SSR page.

CaerusKaru:
> `platform-server` doesn't know in which context the module is being rendered and I don't think it should. It shouldn't matter if we are rendering an app-shell, SSG or SSR page.

You're correct, but that doesn't mean it shouldn't be responsible for marshaling the change into the DOM with consistency and ubiquity. App shell still means there was pre-rendering done, which is useful guidance for us. Universal can then override this string to be more specific based on the use case.

AndrewKushnir:
Hi team! We plan to collect some stats from the HTTPArchive and it'd be great to have an ability to identify sites that used SSR or SSG. This change would be super helpful to achieve this. Just wanted to ask if there are additional discussions/reviews that should happen or we can land the change as is and refactor later if needed? Thank you.

CaerusKaru:
In discussions offline, we realized that to get the coverage that we need on reporting metrics, we need this string to be populated by default in platform-server, and then configurable by runtime (e.g. express-engine, app-shell, etc). Anything less than that would create gaps in reporting for those that roll their own rendering solution without the Angular Universal engines.

alan-agius4:
As @CaerusKaru pointed out this can create gaps for applications not using the NgUniversal but calling the `renderModule` directly. I am still not conceived that given control to the user is the right thing to do here. As we'd rely on the users to set the right thing and it would extended the unnecessary the API for something that the renderer doesn't need.

Maybe something like
```html
<meta property="ng:page-mode" content="SSR/SSG/CSR/Unset" />
```
or
```html
<aio-shell ng-version="13.2.2" ng-page-mode="SSR/SSG/Unset">
```

That said, at the moment from Web archive we can identify if an Application is using SSG.

It's also important to point that in an application you can have a hybrid approach were certain pages are SSG and others are SSR

AndrewKushnir:
Thanks for the replies!

I noticed that there is a new PR created by @CaerusKaru that adds a special token: https://github.com/angular/angular/pull/46887. @alan-agius4 I was wondering if it might be ok to have that special token created in https://github.com/angular/angular/pull/46887, but keep it private (by prefixing with `ɵ`, i.e. `ɵSERVER_CONTEXT`), so it can be leveraged in NgUniversal without declaring it as a public API. WDYT?

alan-agius4:
If the API is private, there is no real reason why the API should be provided by the platform-server IMHO. 

We have control over the index.html, and thus the caller of `renderModule`/`renderApplication` can add whatever context they need.

The server context in this case is used only to collect stats during runtime.

mgechev (approved):

AndrewKushnir:
@alan-agius4 we've landed a PR on the `platform-server` side to be able to pass a `document` reference: https://github.com/angular/angular/pull/47032, but it'd be released in a few weeks from now and would require some extra changes on the NgUniversal side (to parse the DOM). I was wondering if we can update this PR and use a similar approach (string replacement) to add a `<meta>` tag before a closing `</head>` tag for now, WDYT?

alan-agius4:
yeah we can do that as an interim solution.

alan-agius4:
Superseded by https://github.com/angular/universal/pull/2846

PR #2683, opened 2022-06-04
fix(@nguniversal/common): correctly handle headers for jsdom in Clover

hiepxanh:
---
name: ⚙ fix
about: [correctly handle headers for jsdom in Clover
---

## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/angular/universal/blob/main/CONTRIBUTING.md#commit

the server agent is not working in angular. it just default fallback of jsdom

image

the user agent is not corect ly handle by Clover, we should make it correct.

Right now the userAgent is passing in jsdom options which is not correct. We should move it to customResourceLoader  

## What is the new behavior?

the user agent handle correctly as jsdom suggestion. Move userAgent to ResouceLoader as document

image

image

after modification, the userAgent work correctly

image

we just update it as the document

```
[ ] Yes
[x] No
```

## Other information

I'm runing Clover on production app on https://awread.vn. I tested this change on my production and it work very good. I love this engine so much. Thanks for your effort

### DIFF
--- a/modules/common/clover/server/src/custom-resource-loader.ts
+++ b/modules/common/clover/server/src/custom-resource-loader.ts
@@ -12,11 +12,14 @@ import { normalize } from 'path';
 
 export class CustomResourceLoader extends ResourceLoader {
   constructor(
-    private readonly baseUrl: string,
+    readonly headers: Record<string, string | undefined | string[]> | undefined,
     private readonly publicPath: string,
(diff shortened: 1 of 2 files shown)


hiepxanh:
@alan-agius4  can you take a look?

alan-agius4 (approved):
LGTM

PR #2694, opened 2022-06-13
feat(@nguniversal/common): update clover docs and add more jsdom option

hiepxanh:
---
name: ⚙ feat
about:  update clover docs and add more jsdom option
---
In this PR:
- fix error: `Error: self signed certificate` on JSDOM by add two jsdom option: `ssl` and `proxy`
- update document: fix the example url to fit with typescript server (currently format is despricated)
- add document: add more detail about available render option

## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/angular/universal/blob/main/CONTRIBUTING.md#commit

Today, I'm working with clover and need test SSL server to check authentication. And I got error  `Error: self signed certificate`. So I debug about 6 hours and find the true problem come from JSdom. I try to patch it as the document and it work.

image

So I decide to add two option and add document about it.

## What is the new behavior?

add new option for `ssl` and `proxy`

```
[ ] Yes
[x] No
```

## Other information

I'm runing Clover on production app on https://awread.vn. I tested this change on my production and it work very good. I love this engine so much. Thanks for your effort


hiepxanh:
there is merge conflict, I just want to rewrite these file to be more clear, but it just conflict. If I resolve it, it create an auto commit which is fail by commit lint, and I unable to fix it. :(

modules/common/clover/README.md
```diff
@@ -118,3 +113,13 @@ app.listen(PORT, () => {
...
 ```
+
+### Available render options:
+
+- `publicPath`: The public path of the application.
+- `url`: The URL of the application.
+- `headers`: The headers of the request.
+- `inlineCriticalCss`: Whether to inline critical CSS.
+- `htmlFilename`: The filename of the HTML file.
+- `proxy`: The proxy server to use.
```
alan-agius4, line 124: Why is the proxy option needed?

modules/common/clover/server/src/server-engine.ts
```diff
@@ -51,10 +53,12 @@ export class Engine {
...
     const customResourceLoader = new CustomResourceLoader(
-      options.headers,
-      options.publicPath,
-      origin,
       this.resourceLoaderCache,
+      origin,
+      options.publicPath,
+      options.headers,
+      options.proxy,
+      options.strictSSL,
```
alan-agius4, line 61: I wonder if we should test this to false during development and true during production

PR #2697, opened 2022-06-15
Bump version to "v14.0.1" with changelog.

alan-agius4:


CaerusKaru (approved):
Special thanks to @alan-agius4 and @hiepxanh 🎉

PR #2705, opened 2022-06-22
build: update ng-dev and account for ESM-config changes

devversion:


alan-agius4 (approved):
LGTM

devversion:
I probably won't be able to get to that. We should figure out why it conflicts here. We just branched off recently and everything should still apply cleanly

PR #2777, opened 2022-08-12
build: WIP run e2e tests with bazel

gregmagolan:
Just getting started here and doing some experimentation. Will keep as DRAFT for a while.


gregmagolan:
symlinked node_modules tree causing isues with universal code. going to try a different approach for this.

PR #2787, opened 2022-08-18
build: run integration tests with bazel

gregmagolan:
Integration tests passing locally and on CI.

```
//integration/clover:test                                                PASSED in 84.4s
//integration/common-engine:test                                         PASSED in 147.5s
//integration/express-engine-ivy:test                                    PASSED in 165.1s
//integration/express-engine-ivy-hybrid:test                             PASSED in 172.3s
//modules/builders:unit_test                                             PASSED in 247.9s
//modules/common:unit_test                                               PASSED in 0.8s
//modules/common/schematics:unit_tests                                   PASSED in 3.7s
//modules/express-engine:unit_test                                       PASSED in 1.5s
//modules/express-engine/schematics:unit_tests                           PASSED in 4.0s
```

I haven't removed the legacy integration test job yet in this PR which is still passing as the tests are configured to run both under bazel and outside of bazel. The trouble with adding the integration tests to the existing bazel job, however, is that it might make the feedback for unit test failures much slower for some PRs. Best option might be to run the bazel integration tests in a step after the unit tests pass similar to how the legacy integration test step is run currently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,10 +74,12 @@ jobs:
     resource_class: xlarge
     steps:
       - custom_attach_workspace
+      - browser-tools/install-chrome
       - copy-bazel-config
+      # Install rsync which is used by the integration test launcher bash script
+      - run: sudo apt-get update -y && sudo apt-get install -y rsync
       - run: yarn bazel test //...
       - run: scripts/build-modules-dist.sh
-
       - save_cache:
           key: *cache_key
           paths:
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -39,12 +39,24 @@ load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 
 yarn_install(
     name = "npm",
+    all_node_modules_target_name = "node_modules_all",
     # Currently disabled due to:
     #  1. Incompatibilites with the `ts_library` rule.
     exports_directories_only = False,
+    manual_build_file_contents = """\
+# Used by integration tests
+filegroup(
+    name = "node_modules_files",
+    srcs = ["node_modules"],
+)
+""",
     package_json = "//:package.json",
     # We prefer to symlink the `node_modules` to only maintain a single install.
     # See https://github.com/angular/dev-infra/pull/446#issuecomment-1059820287 for details.
     symlink_node_modules = True,
     yarn_lock = "//:yarn.lock",
 )
+
+load("@npm//@angular/build-tooling/bazel/browsers:browser_repositories.bzl", "browser_repositories")
+
+browser_repositories()
--- a/integration/clover/BUILD.bazel
+++ b/integration/clover/BUILD.bazel
@@ -1,0 +1,17 @@
+sh_test(
+    name = "test",
+    srcs = ["test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = glob(
+        ["**"],
+        exclude = ["BUILD.bazel"],
+    ) + [
+        "@npm//:node_modules_files",
+        "//modules/builders:npm_package",
+        "//modules/common:npm_package_runfiles",
+        "//:node_files",
+    ],
+    # e2e test runner creates a tmpdir sandbox itself so no point in double sandboxing
+    tags = ["local"],
+    toolchains = ["@nodejs_toolchains//:resolved_toolchain"],
+)
--- a/integration/clover/test.sh
+++ b/integration/clover/test.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset
+
+# put node on the path
+node_path=$(dirname "$1")
+if [[ "$node_path" == external/* ]]; then
+    node_path="${node_path:9}"
+fi
+PATH="$PWD/../$node_path:$PATH"
+
+tmpdir=$(mktemp -d)
+
+echo "Copying e2e test files to tmpdir sandbox $tmpdir"
+mkdir -p "$tmpdir/integration/clover"
+rsync -a -L integration/clover/ "$tmpdir/integration/clover" --exclude node_modules
+rsync -a ../npm/node_modules/ "$tmpdir/node_modules"
+mkdir -p "$tmpdir/dist/modules-dist"
+rsync -a -L modules/common/npm_package/ "$tmpdir/dist/modules-dist/common" --exclude node_modules
+rsync -a -L modules/builders/npm_package/ "$tmpdir/dist/modules-dist/builders" --exclude node_modules
+
+cd "$tmpdir/integration/clover"
+
+echo "Running e2e test yarn install"
+yarn --mutex network cache clean @nguniversal/common
+yarn --mutex network cache clean @nguniversal/builders
+yarn --mutex network install
+
+echo "Running e2e test"
+yarn test
--- a/integration/common-engine/BUILD.bazel
+++ b/integration/common-engine/BUILD.bazel
@@ -1,0 +1,48 @@
+sh_test(
+    name = "test",
+    srcs = ["test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = glob(
+        ["**"],
+        exclude = [
+            "BUILD.bazel",
+            "node_modules",
+            "dist",
+        ],
+    ) + [
+        "@npm//:node_modules_files",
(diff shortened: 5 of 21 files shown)


patches/@angular+dev-infra-private+0.0.0-e5a5a4c9edf515bcafb9b72a90fe2ad3e811df1d.patch
```diff
@@ -0,0 +1,71 @@
+diff --git a/node_modules/@angular/dev-infra-private/bazel/browsers/browser_archive_repo.bzl b/node_modules/@angular/dev-infra-private/bazel/browsers/browser_archive_repo.bzl
```
gregmagolan, line 1: should upstream this patch to dev-infra-private

package.json
```diff
@@ -104,5 +105,8 @@
     "typescript": "4.8.1-rc",
     "xhr2": "^0.2.0",
     "zone.js": "^0.11.0"
+  },
+  "dependencies": {
+    "patch-package": "^6.4.7"
```
gregmagolan, line 110: can be removed once dev-infra-private patch is upstreamed

modules/common/BUILD.bazel
```diff
@@ -36,6 +36,16 @@ ng_package(
     ],
 )
 
+# work-around for ng_package not providing its output in runfiles
```
gregmagolan, line 39: worth fixing in ng_package?

integration/common-engine/BUILD.bazel
```diff
@@ -0,0 +1,30 @@
+sh_test(
+    name = "test",
+    srcs = ["test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = glob(["**"], exclude = ["BUILD.bazel", "node_modules", "dist"]) + [
+        "@npm//:node_modules_files",
+        "//modules/builders:npm_package",
+        "//modules/common:npm_package_runfiles",
+        # This eager toolchain fetching should be cleaned up in the future
+        "@nodejs_darwin_amd64//:node_files",
```
gregmagolan, line 10: should add select statements here so only the toolchains for the host platform are fetched

integration/express-engine-ivy-hybrid/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+    name = "test",
+    srcs = ["test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = glob(["**"], exclude = ["BUILD.bazel", "node_modules", "dist"]) + [
+        "@npm//:node_modules_files",
+        "//modules/builders:npm_package",
+        "//modules/common:npm_package_runfiles",
+        "//modules/express-engine:npm_package_runfiles",
+        # This eager toolchain fetching should be cleaned up in the future
+        "@nodejs_darwin_amd64//:node_files",
```
gregmagolan, line 11: should add select statements here so only the toolchains for the host platform are fetched

integration/clover/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
+sh_test(
+    name = "test",
+    srcs = ["test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = glob(["**"], exclude = ["BUILD.bazel"]) + [
+        "@npm//:node_modules_files",
+        "//modules/builders:npm_package",
+        "//modules/common:npm_package_runfiles",
+        # This eager toolchain fetching should be cleaned up in the future
+        "@nodejs_darwin_amd64//:node_files",
```
gregmagolan, line 10: should add select statements here so only the toolchains for the host platform are fetched

integration/express-engine-ivy/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
...
+    name = "test",
+    srcs = ["test.sh"],
+    args = ["$(NODE_PATH)"],
+    data = glob(["**"], exclude = ["BUILD.bazel", "node_modules", "dist"]) + [
+        "@npm//:node_modules_files",
+        "//modules/builders:npm_package",
+        "//modules/common:npm_package_runfiles",
+        "//modules/express-engine:npm_package_runfiles",
+        # This eager toolchain fetching should be cleaned up in the future
+        "@nodejs_darwin_amd64//:node_files",
```
gregmagolan, line 11: should add select statements here so only the toolchains for the host platform are fetched

integration/clover/test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -o errexit -o nounset
+
+# put node on the path
```
gregmagolan, line 4: is it worth consolidating common bits from the 3 test.sh files to a bash helper library?

.circleci/config.yml
```diff
@@ -74,10 +74,11 @@ jobs:
     resource_class: xlarge
     steps:
       - custom_attach_workspace
+      - browser-tools/install-chrome
```
gregmagolan, line 77: bazel integration use dev-infra-private browser repositories but this install is still needed for shared libs to be installed

.circleci/config.yml
```diff
@@ -74,10 +74,11 @@ jobs:
...
       - copy-bazel-config
+      - run: sudo apt-get update -y && sudo apt-get install -y rsync
```
gregmagolan, line 79: integration test.sh files use rsync to setup the tmp sandbox directory

WORKSPACE
```diff
@@ -47,4 +47,18 @@ yarn_install(
     # See https://github.com/angular/dev-infra/pull/446#issuecomment-1059820287 for details.
     symlink_node_modules = True,
     yarn_lock = "//:yarn.lock",
+    data = [
+        "//:patches/@angular+dev-infra-private+0.0.0-e5a5a4c9edf515bcafb9b72a90fe2ad3e811df1d.patch",
+    ],
+    manual_build_file_contents = """\
+filegroup(
+    name = "node_modules_files",
```
gregmagolan, line 55: used by integration tests

patches/@angular+dev-infra-private+0.0.0-e5a5a4c9edf515bcafb9b72a90fe2ad3e811df1d.patch
alan-agius4: Should or will😀?

integration/clover/test.sh
alan-agius4: Yes, I think so. Although we can do that as a follow up.

patches/@angular+dev-infra-private+0.0.0-e5a5a4c9edf515bcafb9b72a90fe2ad3e811df1d.patch
gregmagolan: Will :) I'll put up a PR

package.json
gregmagolan: done; also put up a pre-factor related to updating to the latest dev tooling releases https://github.com/angular/universal/pull/2839

patches/@angular+dev-infra-private+0.0.0-e5a5a4c9edf515bcafb9b72a90fe2ad3e811df1d.patch
gregmagolan: ✅

alan-agius4:
@gregmagolan can you please rebase.

integration/express-engine-ivy/BUILD.bazel
gregmagolan: done

gregmagolan:
> @gregmagolan can you please rebase.

Done. And cleaned up the node & chromium files so they are selected based on the target platform (assuming the target is the same as execution which should be true since those tests do not run remote)

integration/express-engine-ivy-hybrid/BUILD.bazel
gregmagolan: Done

modules/common/BUILD.bazel
josephperrott: Yeah, we definitely should and then we can fixup this later.

josephperrott (approved):
LGTM

PR #2793, opened 2022-08-26
fix(@nguniversal/common): handle cookies with localhost domain as path

alan-agius4:
`tough-cookie` which is a dependency of JSDOM did a breaking change in version 4.1 which requires special handling for cookies in localhost.

See: https://github.com/salesforce/tough-cookie/issues/246


CaerusKaru (approved):
I wonder if we should allow this to be disabled by users, but ok for now.

PR #2802, opened 2022-09-01
build: update Angular to version 14.2 and TypeScript to 4.8

alan-agius4:


package.json
```diff
@@ -101,7 +101,7 @@
     "ts-node": "10.9.1",
     "tslib": "~2.4.0",
     "tsutils": "^3.0.0",
-    "typescript": "4.7.4",
+    "typescript": "4.8.1-rc",
```
alan-agius4, line 104: Cannot use the stable version yet as this is blocked on a CLI release.

CaerusKaru (approved):

PR #2839, opened 2022-10-05
build: update to @angular/ng-dev package

gregmagolan:
This package replaces @angular/dev-infra-private/ng-dev

Pre-factor to https://github.com/angular/universal/pull/2787


gregmagolan:
Nevermind. This is hitting build failures. I can move forward without this upgrade.

PR #2846, opened 2022-10-11
Add rendering context.

alan-agius4:
See each commit

### DIFF
--- a/modules/builders/src/prerender/worker.ts
+++ b/modules/builders/src/prerender/worker.ts
@@ -6,6 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import type { Type } from '@angular/core';
+import type * as platformServer from '@angular/platform-server';
+import assert from 'assert';
 import * as fs from 'fs';
 import * as path from 'path';
 import { loadEsmModule } from '../utils/utils';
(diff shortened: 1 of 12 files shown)


CaerusKaru (approved):

CaerusKaru (approved):

modules/builders/src/prerender/worker.ts
```diff
@@ -41,28 +44,39 @@ export async function render({
...
-  const { renderModule, AppServerModule } = await import(serverBundlePath);
+  const { AppServerModule, renderModule, ɵSERVER_CONTEXT } = (await import(serverBundlePath)) as {
+    renderModule: typeof platformServer.renderModule | undefined;
+    ɵSERVER_CONTEXT: typeof platformServer.ɵSERVER_CONTEXT | undefined;
+    AppServerModule: Type<unknown> | undefined;
+  };
+
+  assert(renderModule, `renderModule was not exported from: ${serverBundlePath}.`);
+  assert(AppServerModule, `AppServerModule was not exported from: ${serverBundlePath}.`);
+  assert(ɵSERVER_CONTEXT, `ɵSERVER_CONTEXT was not exported from: ${serverBundlePath}.`);
```
Platonn, line 55: Hi @alan-agius4
Are there plans to make `ɵSERVER_CONTEXT` a public token in v15 (without `ɵ` symbol prefix)?
It would be great to consume it "legally" in the server angular app.
If I understand correctly, according to above line with `assert()`, it's also _required_ that the application's code re-exports the token `ɵSERVER_CONTEXT` (e.g. in`main.server.ts`).

```ts
// main.server.ts
export { 
   renderModule,
   ɵSERVER_CONTEXT   // <-------------- this is a new _required_ thing
} from '@angular/platform-server';
export { AppServerModule } from './app/app.server.module';
```

Side note: if it will be required to re-export this token in v15, it would be nice to provide a migration schematics for it in v15.

Platonn: Now I've just created a fresh Angular app with  `@angular/cli@15.0.0-rc.1` and `@nguniversal/express-engine@15.0.0-next.0` and the generated file `main.server.ts` contains only `export { AppServerModule } from './app/app.server.module';`
Perhaps I'm missing something in the internals of the `prerender` builder(?)... how it's possible, that `renderModule` and `ɵSERVER_CONTEXT` are exported from the output `main.js` bundle?

PS. Regardless of the above analysis, I still think, it would be cool to drop `ɵ` prefix from `ɵSERVER_CONTEXT`, if only possible in v15 :)

Platonn: Ok. Now I've figured out the magic that adds the re-exports of `renderModule` and `ɵSERVER_CONTEXT` under the hood! 😅  
Merged PR for v15: https://github.com/angular/universal/pull/2846
In particular those lines: https://github.com/angular/angular-cli/pull/23866/files#diff-e1778745f6c1b0fbef4fe31a8e09d71863363d6980922e3df974c408772025d2R21-R23

PS. IMHO it seems justified to export `SERVER_CONTEXT` the same as `PLATFROM_ID` is exported in the public API.

alan-agius4: The `ɵSERVER_CONTEXT` is meant to be set internally as such it is considered a private API which it's implementation can change in the future. 

Users should not need nor need to set this token.

PR #2990, opened 2023-01-16
build: remove minimatch as it's unused

alan-agius4:


yarn.lock
```diff
@@ -6658,7 +6658,7 @@ minimatch@^3.0.2, minimatch@^3.0.4, minimatch@^3.0.5, minimatch@^3.1.1, minimatc
   dependencies:
     brace-expansion "^1.1.7"
 
-minimatch@^5.0.0, minimatch@^5.0.1:
+minimatch@^5.0.1:
```
CaerusKaru, line 6661: If we're removing the dep and this is still here, does something else depend on it?

CaerusKaru (approved):

yarn.lock
alan-agius4: It means that it's a transitive dependency in which case we don't need to list in the `package.json`

PR #3045, opened 2023-03-13
Standalone rendering

alan-agius4:
See each commit.


alan-agius4:
Blocked on https://github.com/angular/angular-cli/pull/24847

integration/express-engine-standalone/README.md
```diff
@@ -0,0 +1,27 @@
+# universal-integration
+
+This project was generated with [Angular CLI](https://github.com/angular/angular-cli) version 9.0.0-next.4.
```
robertIsaac, line 3: why did you generate it using angular 9?
I can see other changes as well, like tsconfig target ES5, polyfill imported in the old way, tslint being used for linting not eslint

alan-agius4: This is just a test project which is used for testing. Which was copied from existing test to retain the same setup.

CaerusKaru (approved):

PR #3057, opened 2023-03-22
fix(@nguniversal/common): update TransferHttpResponse requiring body and headers

bougwal:
As per the discussion and @alan-agius4's point regarding the `TransferHttpResponse` shape under https://github.com/angular/angular/pull/49509. I suggest to align the status of optional and required keys of `TransferHttpResponse` interface of universal's interceptor to meet Angular's `packages/common/http/src/transfer_cache.ts `counterpart.

### DIFF
--- a/modules/common/clover/src/transfer-http-cache/transfer-http-cache.interceptor.ts
+++ b/modules/common/clover/src/transfer-http-cache/transfer-http-cache.interceptor.ts
@@ -23,8 +23,8 @@ import { filter, take, tap } from 'rxjs/operators';
 type ResponseType = HttpRequest<unknown>['responseType'];
 
 interface TransferHttpResponse {
-  body?: any | null;
-  headers?: Record<string, string[]>;
+  body: any;
+  headers: Record<string, string[]>;
   status?: number;
   statusText?: string;
   url?: string;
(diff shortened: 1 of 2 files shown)


alan-agius4:
@bougwal please sign the CLA.

modules/common/clover/src/transfer-http-cache/transfer-http-cache.interceptor.ts
```diff
@@ -75,7 +75,7 @@ export class TransferHttpCacheInterceptor implements HttpInterceptor {
 
     if (this.transferState.hasKey(storeKey)) {
       // Request found in cache. Respond using it.
-      const response = this.transferState.get(storeKey, {});
+      const response = this.transferState.get(storeKey, { body: {}, headers: {} });
```
alan-agius4, line 78: Use a default value of `null` instead

alan-agius4: I meant     `const response = this.transferState.get(storeKey, null);`

alan-agius4: Yes please do.

alan-agius4 (approved):
LGTM, thanks for the contribution

PR #3072, opened 2023-04-05
fix(@nguniversal/express-engine): update import for template server file

brandonroberts:
The `server.ts` is missing a single tick on the import


alan-agius4 (commented):
Can you please update the commit message to `fix(@nguniversal/express-engine): …`?

brandonroberts:
Yep, done

alan-agius4:
> Yep, done

The scope is still incorrect.

alan-agius4:
Looks like you need to update the unit test 

https://github.com/angular/universal/blob/7b767d2c1514066fee7a77e5a241b92fcd035045/modules/express-engine/schematics/install/index.spec.ts#L112

alan-agius4 (approved):
LGTM, thanks

PR #3081, opened 2023-04-12
fix(@nguniversal/builders): add support to serve service workers

alan-agius4:
Previously, for legacy reasons service workers were not served when using the `ssr-dev-server` builders. This commit enabled serving of service workers as now service workers can be served without valid SSL when served using localhost.


CaerusKaru (approved):
Exciting!

PR #3083, opened 2023-04-12
build: build packages in APF version 16 format

alan-agius4:
This commit update tooling to generate APF version 16 distributables.

### DIFF
--- a/tools/postinstall/patches/@bazel+concatjs+5.8.1.patch
+++ b/tools/postinstall/patches/@bazel+concatjs+5.8.1.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/@bazel/concatjs/internal/build_defs.bzl b/node_modules/@bazel/concatjs/internal/build_defs.bzl
+index 9e5cda6..851c8b7 100755
+--- a/node_modules/@bazel/concatjs/internal/build_defs.bzl
++++ b/node_modules/@bazel/concatjs/internal/build_defs.bzl
+@@ -76,7 +76,7 @@ _TYPESCRIPT_TYPINGS = Label(
+     "//typescript:typescript__typings",
+ )
+ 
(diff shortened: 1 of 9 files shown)


tools/postinstall/patches/@bazel+concatjs+5.8.1.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@bazel/concatjs/internal/build_defs.bzl b/node_modules/@bazel/concatjs/internal/build_defs.bzl
+index 9e5cda6..851c8b7 100755
+--- a/node_modules/@bazel/concatjs/internal/build_defs.bzl
++++ b/node_modules/@bazel/concatjs/internal/build_defs.bzl
+@@ -76,7 +76,7 @@ _TYPESCRIPT_TYPINGS = Label(
+     "//typescript:typescript__typings",
+ )
+ 
+-_TYPESCRIPT_SCRIPT_TARGETS = ["es3", "es5", "es2015", "es2016", "es2017", "es2018", "es2019", "es2020", "esnext"]
++_TYPESCRIPT_SCRIPT_TARGETS = ["es3", "es5", "es2015", "es2016", "es2017", "es2018", "es2019", "es2020", "es2022", "esnext"]
```
devversion, line 10: do you need this here? Should all ts library targets set the `prodmode_target = es2022`

alan-agius4: This is needed as otherwise es2022 will not be supported as it will fail input validation. 

https://github.com/bazelbuild/rules_nodejs/blob/95fcbad358736d5ce93b3623482643c3e79e202c/packages/concatjs/internal/build_defs.bzl#L427

devversion: yeah, but where are you setting the value to es2022?

devversion: Yes, that's what my comment is about. You did not do it for ts library

devversion (approved):

PR #3159, opened 2023-06-01
refactor: deprecate transfer http cache

alan-agius4:
DEPRECATION: `StateTransferInitializerModule`, `TransferHttpCacheInterceptor` and `domContentLoadedFactory` have been deprecated in favor of `provideClientHydration`. See: https://angular.io/api/platform-browser/provideClientHydration


CaerusKaru (approved):
🎉

PR #3209, opened 2023-07-30
docs: update the http tag with https in contributing file

aanchal88:


CaerusKaru (approved):

dgp1130:
Apparently I don't have access to merge to patch so this change only landed on `main`. I think that's probably fine for a markdown change like this, but I'll follow up with Alan to double check branch protection and repo permission settings.

PR #3210, opened 2023-07-30
docs: remove unused methods from lazy component

aanchal88:


alan-agius4:
Thanks for your contribution but seeing this is a test, there is no value in doing this change.

Issue #3241, opened 2023-08-16
"TypeError: link.prev?.setAttribute is not a function" upon rendering any page with "ngCspNonce" set.


majoer:
# 🐞 Bug report

#### What modules are related to this issue?

- [ ] builders
- [x] common
- [ ] express-engine

### Is this a regression?
I don't think so, as the CSP feature was added in 16

### Description

We started using the "ngCspNonce" attribute when we upgraded to Angular 16. The application builds and starts, but upon sending a request to a SSR page we get an "Internal server error" in the browser. The console also displays an error message:
<pre><code>
TypeError: link.prev?.setAttribute is not a function
</code></pre>

The same bug was reported in @angular-devkit/build-angular here:
- https://github.com/angular/angular-cli/issues/25419 
- https://github.com/angular/angular-cli/issues/25575

This is the PR where they fixed it: https://github.com/angular/angular-cli/pull/25584
After this fix was released I expected our app to work, but I found angular universal has its own version of "inline-css-processor.ts" here:
https://github.com/angular/universal/blob/5974f8e769e75023cacbf0cc52f7c1a6ae534f45/modules/common/tools/src/inline-css-processor.ts#L157C21-L157C21

I believe the fix should be applied to Angular Universal as well.

## 🔬 Minimal Reproduction

1. Add "ngCspNonce" to the app-root as explained in: https://angular.io/guide/security#content-security-policy
2. Add some external style-sheets as links in index.html.
3. Set `inlineCritical": true` in angular.json
4. build and run with SSR

Here is the repo i made for the issue I reported in angular-cli. The same setup should yield the same problem in angular universal: https://github.com/majoer/angular-csp-link-bug

## 🔥 Exception or Error

<pre><code>
TypeError: link.prev?.setAttribute is not a function
</code></pre>

## 🌍 Your Environment

<pre><code>
Angular CLI: 16.2.0
Node: 16.17.0
Package Manager: npm 8.15.0
OS: linux x64

 

Angular: 16.2.0
... animations, cdk, cli, common, compiler, compiler-cli, core
... forms, language-service, localize, material
... material-moment-adapter, platform-browser
... platform-browser-dynamic, platform-server, router

 

Package                         Version
---------------------------------------------------------
@angular-devkit/architect       0.1602.0
@angular-devkit/build-angular   16.2.0
@angular-devkit/core            16.2.0
@angular-devkit/schematics      16.2.0
@nguniversal/builders           16.2.0
[...]

PR #3243, opened 2023-08-17
fix(@nguniversal/common): error during critical CSS inlining for external stylesheets

alan-agius4:
The problem with the previous approach is that it assumes that a `link` tag processed by Critters will be preceded by a `style` tag that was inserted by Critters. This assumption might not necessarily be true if Critters is unable to resolve the content of the linked styles.

Closes: #3241

### DIFF
--- a/modules/common/tools/src/inline-css-processor.ts
+++ b/modules/common/tools/src/inline-css-processor.ts
@@ -154,7 +154,15 @@ class CrittersExtended extends Critters {
         this.conditionallyInsertCspLoadingScript(document, cspNonce);
       }
 
-      link.prev?.setAttribute('nonce', cspNonce);
+      // Ideally we would hook in at the time Critters inserts the `style` tags, but there isn't
+      // a way of doing that at the moment so we fall back to doing it any time a `link` tag is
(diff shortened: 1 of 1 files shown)


modules/common/tools/src/inline-css-processor.ts
```diff
@@ -154,7 +154,15 @@ class CrittersExtended extends Critters {
         this.conditionallyInsertCspLoadingScript(document, cspNonce);
       }
 
-      link.prev?.setAttribute('nonce', cspNonce);
+      // Ideally we would hook in at the time Critters inserts the `style` tags, but there isn't
+      // a way of doing that at the moment so we fall back to doing it any time a `link` tag is
+      // inserted. We mitigate it by only iterating the direct children of the `<head>` which
+      // should be pretty shallow.
+      document.head.children.forEach((child) => {
```
CaerusKaru, line 161: This makes me wonder if we should be using a DOM query/find API, but I'm ok with a limited hand-rolled version since we know what we want.

alan-agius4: This DOM parser does not support queries.

PR #3246, opened 2023-08-25
docs: removed unused ngOnInit and constructor from lazy module

alkavats1:


alan-agius4:
Thanks for this PR, but the done changes are in a test project and thus offer no benefit.

PR #3255, opened 2023-10-10
docs: update README

dgp1130:
Updates the root README to better explain repository status in Angular v17 and beyond. The v17 blog post, livestream, and other DevRel content should elaborate more, but hopefully this is good enough to communicate project status to contributors.

CC @mgechev in case he has any thoughts on the best of way of communicating this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,24 @@
 # Angular Universal
 
-## This repository has been moved to Angular CLI repo
+Angular Universal was a project to expand on the core APIs from Angular
+(platform-server) to enable developers to do server side rendering of Angular
+applications in a variety of scenarios.
 
-You can find the Angular CLI repo [here](https://github.com/angular/angular-cli).
+## This repository has been merged into the Angular CLI repo
+
+In version 17, Universal has been moved into the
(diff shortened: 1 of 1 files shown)


CaerusKaru (approved):

README.md
```diff
@@ -1,5 +1,25 @@
...
 
-## This repository has been moved to Angular CLI repo
+Angular Universal was a project to expand on the core APIs from Angular
+(platform-server) to enable developers to do server side rendering of Angular
+applications in a variety of scenarios.
 
-You can find the Angular CLI repo [here](https://github.com/angular/angular-cli).
+## This repository has been merged into the Angular CLI repo
+
+While Universal has been a key part of Angular's server-side rendering and
```
MarkTechson, line 9: >> While Universal has been a key part of Angular's server-side rendering and
build-time prerendering story, in version 17 its code has been moved into the
[Angular CLI repo

I would consider changing this to be more brief so that the point does not get obscured.
In version 17 its code has been moved into the
[Angular CLI repo

dgp1130: Good point. Updated to:

> In version 17, Universal has been moved into the Angular CLI repo.

alan-agius4 (approved):

